Issue #12, opened 2016-08-30
Print the version correctly


dvitas:
A) Implement "-V" option in mstpd;
B) Print out the real version both in mstpd and mstpctl. For this VERSION define from config.h should be used.

Also it would be nice to use the output of the following command:
<pre>
  git log --pretty=format:'%h' -n 1
</pre>
as the BUILD number when showing version

PR #14, opened 2016-08-30
Print real version both in mstpd and mstpctl

dvitas:
Resolves issue #12

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,8 +8,10 @@ mstpd_SOURCES = \
 mstpctl_SOURCES = \
 	ctl_main.c ctl_socket_client.c
 
+BUILDDEF = $(shell git log --pretty=format:'%h' -n 1)
+
 mstpd_CFLAGS = \
-	-Os -Wall -Werror -D_REENTRANT -D__LINUX__ -DVERSION=$(version) -I. \
+	-Os -Wall -Werror -D_REENTRANT -D__LINUX__ -DVERSION=$(VERSION) -DBUILD="$(BUILDDEF)" -I. \
 	-D_GNU_SOURCE -D__LIBC_HAS_VERSIONSORT__
 if ENABLE_DEVEL
   mstpd_CFLAGS += -g3 -O0
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -1632,9 +1632,6 @@ static void help(void)
     command_helpall();
 }
 
-#define PACKAGE_VERSION2(v, b) "mstp, " #v "-" #b
-#define PACKAGE_VERSION(v, b) PACKAGE_VERSION2(v, b)
-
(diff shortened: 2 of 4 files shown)


dvitas:
Workflow suggestion:
I don't think that we really need PRs for such a small changes. So I suggest the following:
 - keep the master branch more or less stable.
 - create "development" branch and push such a small changes to that branch directly. When we feel that code in the "development" branch is stable enough and has enough significant changes - we merge it ti the master branch and place new version tag.
 - Only lengthy or controversial changes should go in as PRs. Or changes from third parties.

Thoughts?

vivien:
Hi Vitalii,

> Workflow suggestion:
> I don't think that we really need PRs for such a small changes. So I suggest the following:
> 
>    keep the master branch more or less stable.
>    create "development" branch and push such a small changes to that branch directly. When we feel that code in the "development" branch is stable enough and has enough significant changes - we merge it into the master branch and mark it with the new version tag.
>    Only lengthy or controversial changes should go in as PRs. Or changes from third parties.
> 
> Thoughts?

I think we should merge single-commit PRs and trivial PRs manually, to keep linear history. However merging a well formatted feature PR (i.e. clicking the Github merge button) is totally ok (given a correct branch name).

About the branching model, we might want to keep only the master branch, and eventually one branch per release where fixes can be cherry-picked to. That would avoid merges from/to a dev branch. I'm comparing a few projects at the moment to see what will fit mstpd most.

dvitas:
Good, let's go with the manual merge model for a while, we'll see how it works.

Any comments on the PR itself?

commodo:
ack from me for this PR/change

regarding workflow model(s): I'm fine with whichever is decided :)
i usually don't have strong preference for this
i've seen Github projects use PRs used for single commits, so it's not uncommon;

Github's PR model makes git history a bit messy.
Would be nice if they would implement some patch-application model for small commits like this.

vivien:
> Good, let's go with the manual merge model for a while, we'll see how it works.

Perfect. I usually prefer not to have merge commit *unless* it is relevant, i.e. grouping several commits adding a new feature, for the simple reason that a merge commit can be reverted. If several commits are not logically dependent and can be reverted individually, there is no need to merge them with `--no-ff`, à la Github. (It is just my opinion here, feel free to criticize)

> Github's PR model makes git history a bit messy.
> Would be nice if they would implement some patch-application model for small commits like this.

True. I've seen they added the setting to support `git merge --squash`. I don't see why they don't add such support for `--ff-only`, which would prevent a branch from being merged if it isn't up-to-date with master... Anyway :)

Thanks Vitalii for the commit, I've pushed it manually with a small reword to respect [commit keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) and the Git preference for a second empty line.

Closing this issue.

Issue #11, opened 2016-08-30
Detect __LIBC_HAS_VERSIONSORT__ during configure stage


dvitas:
Currently this define is always defined, but it is there not without a reason, obviously there were a system where libc lacked the versionsort function. As we have now the real configure stage, it is only natural to really detect if libc has versionsort during this stage.

Not of the highest priority though. Also it would be good to check the detection on the real system with this kind of legacy libc.

commodo:
if no objections, i'll try to implement this

commodo:
according to glibc's changelog ([this older one](https://sourceware.org/git/?p=glibc.git;a=blob;f=ChangeLog.8;h=52b85a627091757f17056be0ff340e3e305b0cf1;hb=HEAD#l1113) ), versionsort should be in glibc since 1998 (or so)

uClibc implements it: https://git.uclibc.org/uClibc/tree/libc/misc/dirent/versionsort.c
musl libc implements it: https://git.musl-libc.org/cgit/musl/tree/src/dirent/versionsort.c

@dvitas : any recollection about why it was added here https://github.com/mstpd/mstpd/commit/1e16013cc061f64fe4793290d5e17381dc7b84c6 ?
maybe just `_GNU_SOURCE ` being defined is sufficient to have this enabled

so, we could just remove that define and stick to `versionsort` directly ?

commodo:
i'm wondering if the case is some old Solaris systems;
if that's the case, would still we want to address that case ?

dvitas:
If I recall it correctly it was some legacy embedded system using really old uClibc.
I think you are right and we can drop this define and use versionsort unconditionally and fix it only if it is broken for some user of mstpd.
"don't fix it if it's not broken"

PR #15, opened 2016-09-02
mstpd: use 'versionsort' by default

commodo:
According to glibc's changelog ([this older one](https://sourceware.org/git/?p=glibc.git;a=blob;f=ChangeLog.8;h=52b85a627091757f17056be0ff340e3e305b0cf1;hb=HEAD#l1113) ),
versionsort should be in glibc since 1998 (or so)

uClibc implements it: https://git.uclibc.org/uClibc/tree/libc/misc/dirent/versionsort.c
musl libc implements it: https://git.musl-libc.org/cgit/musl/tree/src/dirent/versionsort.c

Unless, we'll get a report for another libc that may not
implement it, we'll stick with this just being available by default.

Resolves: https://github.com/mstpd/mstpd/issues/11

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,7 +12,7 @@ BUILDDEF = $(shell git log --pretty=format:'%h' -n 1)
 
 mstpd_CFLAGS = \
 	-Os -Wall -Werror -D_REENTRANT -D__LINUX__ -DVERSION=$(VERSION) -DBUILD="$(BUILDDEF)" -I. \
-	-D_GNU_SOURCE -D__LIBC_HAS_VERSIONSORT__
+	-D_GNU_SOURCE
 if ENABLE_DEVEL
   mstpd_CFLAGS += -g3 -O0
 endif
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -32,12 +32,6 @@
 #include "ctl_socket_client.h"
 #include "log.h"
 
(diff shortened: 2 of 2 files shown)


commodo:
PR is for review
if all good, I can push the change directly to master ; 
or, if something changed workflow-wise, please update me :)

dvitas:
All good, ACK

commodo:
Fixed via https://github.com/mstpd/mstpd/commit/8fcb118d0e88087103662c059e67593c6e53f365

vivien:
Hi @commodo, you can push it. Quick nitpicks though:

* ideally we don't want to push our own PRs, it is preferable to have someone else doing it
* the Signed-off-by tag is not necessary, until we stipulate a proper contract to apply against it

Again, these are minors concerns (best practices I'd say) and could be worked around sometimes.

commodo:
ack on first bullet point
regarding the `Signed-off-by` tag, for me it's a habit when upstreaming ; it's usually required on other projects, so no strong preference for using it; as long a decision is made on this

thanks :)

vivien:
Yes, I do have the same habits as your for the same reasons ;-)
It was just a friendly reminder for us to know that this tag has no value until we mention a related developer contract.
Given that, I don't mind if we add it or not at the moment.

Issue #3, opened 2016-08-30
Fix time glitches when system time changes (e.g. by NTP)


dvitas:
Substitute gettimeofday() with the clock_gettime(CLOCK_MONOTONIC, ) for the systems where this functionality is present. I think one should check _POSIX_MONOTONIC_CLOCK macro for availability of it.

commodo:
if no objections, i'll try to implement this

PR #16, opened 2016-09-05
mstpd: add support for clock_gettime(CLOCK_MONOTONIC, )

commodo:
Resolves https://github.com/mstpd/mstpd/issues/3


vivien:
Looks good, thanks!
Note: we might consider a `compat` folder and header in the future if more compatibility code is needed.

commodo:
makes sense about the `compat` folder
i also had the thought about doing it now, but was undecided, since it's only 1 compat element right now
but, when the second appears, then definitely

PR #22, opened 2016-10-14
mstpd: add minimal integration with travis-ci.org

commodo:
Seems to be popular around Github.
So, why not have it ?

We could use this as a validation tool for PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ mstpd
 *.swp
 
 # autotools files
+/m4/lt*.m4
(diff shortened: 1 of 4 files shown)


commodo:
I forgot to add ; after adding these patches, we'll need to enable the build from Travis.
Hmm, actually, I should go ahead and do that now, if no objections.
I think anyone in the group can edit that, when logging [with Github account] in Travis.

commodo:
w00h00 : build passing tag FTW

PR #33, opened 2017-02-03
Disable ports on exit  and  handle more signals

fxkr:
Just one pull request since these commits depend on each other since they both deal with signals.


commodo:
thanks

PR #34, opened 2017-03-01
bridge_track: add verbose debug messages in bridge_bpdu_rcv()

commodo:
I got these messages due to a switch + phy driver bug.

But it makes sense to have these a bit more explicit for
future cases.


vivien:
These what-seem-to-be extra commas in TST andTSTM macros are quite disturbing, but this compiles well, thanks!

PR #38, opened 2017-03-29
bridge-stp.in: add a sleep to the mstpctl background workaround

dvlemplgk:
It happens that the background job to add the bridge running to early before the parent shell actually exits. The msptctl call is failing then and STP mode is turned off.
Add a sleep 1 as a workaround.
(Happened to me almost every time during booting of a device running mstpd on OpenWrt.)

### DIFF
--- a/bridge-stp.in
+++ b/bridge-stp.in
@@ -127,7 +127,7 @@ case "$action" in
             # then dying before or when mstpctl connects to it.  To avoid that
             # possibility, we instead simply turn STP off if `mstpctl addbridge`
             # fails.
-            ( "$mstpctl" addbridge "$bridge" || brctl stp "$bridge" off ) &
+            ( sleep 1 ; "$mstpctl" addbridge "$bridge" || brctl stp "$bridge" off ) &
             exit 0
         fi
 


commodo:
ack from me
@PaulSD any input ?

PaulSD:
+1

Should probably have a space before the semicolon to keep the formatting consistent with other parts of the shell scripts in mstpd.

PaulSD:
Does 'sleep .1' fix the problem? A full second is probably an excessive delay - could probably make this much shorter and still fix the issue.

commodo:
i'd go for 1 second ; maybe to account for the slowness of some embedded devices
i do agree with the space for the semi-colon

@dvlemplgk 
could you add a space for the `sleep 1;`  part ?

dvlemplgk:
Yes it did fix the issue. I don't know about a way to sleep for less than 1 s in user space. It would be nice if a subshell could wait for the parent shell to exit somehow, but I couldn't find such a solution.

Space added.

commodo:
@dvlemplgk thanks :)

PaulSD:
Just for reference: On my Ubuntu/RHEL systems, the Linux kernel seems to lock a mutex, and the mstpctl call waits on this mutex for some period of time before failing, so no sleep is needed ... mstpctl simply hangs until bridge-stp exits and releases the mutex, then it continues running normally.  There must be something odd that OpenWRT is doing differently than other Linux systems.

PR #40, opened 2017-05-13
Various debian packaging fixes and improvements

bluca:
See individual commit messages for details.

All Lintian errors are gone, and the only remaining warnings are about missing manpages fro bridge-stp, mstp_restart and mstpd (would be great to have them added), and for missing +x on mstpctl-utils-functions.sh. The latter is a library to be sourced, so perhaps the shebang at the top could be removed?


debian/mstpd.lintian-overrides
```diff
@@ -0,0 +1 @@
+mstpd: read-in-maintainer-script
```
commodo: hmmm ; it may be interesting to try to add this to our `.travis.yml`
i'll take a look

commodo:
[came back from the weekend]

@bluca thanks for the contribution :)
looks good from my side

regarding the shebang in https://github.com/mstpd/mstpd/blob/master/utils/mstpctl-utils-functions.sh : that one was added because of the `shellcheck` check via this commit https://github.com/mstpd/mstpd/commit/6eccfb9f50b5ccf260536e0111804845a00b2cae
i have no strong opinion about removing it ;
i think it comes down to a negotiation here, about whether to remove it , or make shellcheck ignore it ;
personally, i'm leaning towards leaving it, since `shellcheck` can know which level of shell-isms to allow/check for

i'll make an effort to integrate some more CI via `.travis.yml` for the Debian stuff & maybe also add Lintian as a check in the CI [if possible] ; then we can see how to handle some of the other warnings I guess

@PaulSD do you also want to take a look ?

maybe we could also see about making a release tarball

if no objections until Wed [17th of May], I'll merge this

thanks
Alex

bluca:
If you would like to add package builds&tests to Travis, I would recommend looking at this: http://travis.debian.net/

Regarding the script, if shellcheck suggested that change then it's probably a good idea to leave it in. Another Lintian override could be easily added instead. Given it's a function "library", I don't think it would make much sense to mark it as executable.

commodo:
interesting ; will check it out [ http://travis.debian.net/ ] 
ack to make the `mstpctl-utils-functions.sh` file non-executable

commodo:
it's Wed here in in my TZ
no objections so far ;
if any come up, we'll address them in later commits

@bluca thanks

commodo:
according to https://github.com/mstpd/mstpd/blob/master/Makefile.am#L68
`mstpctl-utils-functions.sh` is already installed as non-executable
should have checked earlier

bluca:
Thanks!

```
according to https://github.com/mstpd/mstpd/blob/master/Makefile.am#L68
mstpctl-utils-functions.sh is already installed as non-executable
should have checked earlier
```

Yes - sorry if I wasn't clear, that is what Lintian is complaining about: .sh extension + shebang at the top, but NOT executable:

```
dpkg-buildpackage: info: full upload; Debian-native package (full source is included)
Now running lintian...
W: mstpd: binary-without-manpage sbin/bridge-stp
W: mstpd: binary-without-manpage sbin/mstp_restart
W: mstpd: binary-without-manpage sbin/mstpd
W: mstpd: script-not-executable lib/mstpctl-utils/mstpctl-utils-functions.sh
```

I'll open a new PR to override this warning since we agree it's best to leave it as-is

PaulSD:
Sorry, was busy this week and just saw this.  LGTM, except the prerm script should really use pidof to determine the PID, similar to commit b8a351d in bridge-stp.in.  (This is my fault - I apparently forgot to update the prerm script to use pidof when I updated bridge-stp.in to use pidof.)  I've opened PR #43 to fix that.
Thanks!

commodo:
Thanks for taking the time to review this and the PR regarding pidof

PR #41, opened 2017-05-17
debian: add override for script-not-executable LW

bluca:
utils/mstpctl-utils-functions.sh is a bash library, so it is not
supposed to be ran individually. It has a shebang as it was
recommended by shellcheck, so Lintian generates a false positive.
Override it.


commodo:
ah ; ok

thanks for the clarification & PR

bluca:
Thanks!

PR #42, opened 2017-05-17
build,travis: add support for debian's travis script

commodo:
Adapted from: http://travis.debian.net/

@bluca

https://travis-ci.org/commodo/mstpd/builds/233154741

AFAICT, the build looks passing ; 
tho, the lintian checks/warnings did not produce a build failure
not sure if they should
```
warning: the authors of lintian do not recommend running it with root privileges!
W: mstpd: binary-without-manpage sbin/bridge-stp
W: mstpd: binary-without-manpage sbin/mstp_restart
W: mstpd: binary-without-manpage sbin/mstpd
W: mstpd: script-not-executable lib/mstpctl-utils/mstpctl-utils-functions.sh
```

I don't get why the warning `warning: the authors of lintian do not recommend running it with root privileges!`
maybe that's caused by the fact that it's run in docker

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,18 @@ git:
 language: c
 os: linux
 compiler: gcc
+services:
+  - docker
 addons:
   apt:
     sources:
     - debian-sid
     packages:
     - shellcheck
+env:
+  - TRAVIS_DEBIAN_DISTRIBUTION="${TRAVIS_DEBIAN_DISTRIBUTION:-experimental}"
 script:
   - set -e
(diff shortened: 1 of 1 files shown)


bluca:
Nice!

- Lintian _warnings_ are not supposed to fail a build, so the behaviour is correct
- There are conflicting opinions on whether Lintian _errors_ should fail the build - the Debian build infrastructure does not, but a CI can decide to be stricter
- Yes the "with root" warning is just because normally Debian Developers should not run builds as root for security (and machine stability) purposes, but this are throw-away instances so it's safe to ignore

commodo:
if no objections
i'll merge this in 1-2 days

PR #43, opened 2017-05-20
debian/mstpd.prerm: use pidof to determine mstpd PID

PaulSD:
See b8a351de38dd5fe476527c66179bc7109e57e9d7


commodo:
Looks good.
Thanks :)

PR #45, opened 2017-06-06
bridge-stp.in: test config_cmd for executability

commodo:
A re-spin of https://github.com/mstpd/mstpd/pull/37.

This change checks if the config_cmd is executable,
or checks if the command is in the PATH env var.

I referenced the `type` command from an OpenWrt script.

Hopefully that would ensure that it's portable enough
across other shells as well.
Seems to work on bash & zsh too.


commodo:
@PaulSD 
@dvlemplgk 

will merge in 2 days if no objections

dvlemplgk:
ack

PaulSD:
+1

commodo:
thanks for the quick ack :)
will merge

Issue #39, opened 2017-03-30
Adding a bridge by 'bridge-stp start' vs. 'bridge-stp restart(_config)'


dvlemplgk:
There are two ways to add bridges to mstpd
1. bridge-stp start
In this case the bridge is just added by calling mstpctl addbridge.
2. bridge-stp restart(_config)
In this case bridges are added **and** additionally configured by calling $config_cmd.

Shouldn't bridge-stp start configure the bridge after adding it? Or does debian handle the configuration itself during initial bridge setup?

commodo:
@PaulSD ? :)

PaulSD:
Yes, Debian/Ubuntu handle the configuration themselves during initial bridge setup.  Basically, the utils/ifupdown.sh.in script is called automatically by ifup, and it calls brctl addbr, which calls bridge-stp start, then after that returns, the ifupdown.sh.in script finishes the configuration.

For RedHat, utils/mstp_config_bridge is called by the /etc/sysconfig/network-scripts/ifcfg-* files (as described in the comments in /etc/sysconfig/network-scripts/bridge-stp), and mstp_config_bridge calls the ifupdown.sh.in script, which calls brctl addbr, which calls bridge-stp start, then after that returns, the ifupdown.sh.in script finishes the configuration.

In general, the idea is: 'bridge-stp start' should never be called directly by a user. To bring the bridge up the first time, mstp_config_bridge or ifupdown.sh should always be called, and that will indirectly call 'bridge-stp start' and then configure the bridge, so there is no need for 'bridge-stp start' to configure the bridge itself.  However, mstp_config_bridge/ifupdown.sh are not inherently involved when restarting mstpd, so 'bridge-stp restart_config' needs to call them explicitly to reconfigure each bridge after restarting mstpd, and this is where $config_cmd comes in.

PaulSD:
I think this can be closed?

commodo:
fine from me

PR #48, opened 2017-10-07
Race condition fix and script comment improvements

PaulSD:
Fixes a race condition introduced by https://github.com/mstpd/mstpd/pull/38 and adds lots more comments to explain the scripts relationships/usage (which should help answer questions like https://github.com/mstpd/mstpd/issues/39).


commodo:
Thanks.

I'll merge this even i f Travis fails.
Seems the compiler changed to GCC 7.

I'll take care of that in the next days

PR #50, opened 2017-11-13
ctl_main: fix truncation warnings/errors caught by GCC 7

commodo:
GCC 7 caught this:
```
ctl_main.c: In function 'isbridge':
ctl_main.c:283:25: error: '%s' directive output may be truncated writing up to 255 bytes into a region of size 241 [-Werror=format-truncation=]
^

ctl_main.c:293:36: note: in expansion of macro 'SYSFS_CLASS_NET'
snprintf(path, SYSFS_PATH_MAX, SYSFS_CLASS_NET "/%s/bridge",
^~~~~~~~~~~~~~~

ctl_main.c:705:50: note: format string is defined here
snprintf(buf, sizeof(buf), SYSFS_CLASS_NET "/%s/brif", br_ifname);
^~
In file included from /usr/include/stdio.h:938:0,
from log.h:30,
from ctl_main.c:35:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:64:10: note: '__builtin___snprintf_chk' output between 23 and 278 bytes into a destination of size 256
return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
__bos (__s), __fmt, __va_arg_pack ());
```

### DIFF
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -290,7 +290,9 @@ static int isbridge(const struct dirent *entry)
     bool result;
     struct stat st;
 
-    snprintf(path, SYSFS_PATH_MAX, SYSFS_CLASS_NET "/%s/bridge",
+    /* strlen(SYSFS_CLASS_NET) + strlen("/%.230s/bridge") must be < SYSFS_PATH_MAX
+       to prevent string truncation ; gcc7's fortify headers complain about that */
+    snprintf(path, SYSFS_PATH_MAX, SYSFS_CLASS_NET "/%.230s/bridge",
              entry->d_name);
     save_errno = errno;
     result = (0 == stat(path, &st)) && S_ISDIR(st.st_mode);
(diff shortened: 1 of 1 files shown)


PaulSD:
I think there should be a comment explaining why "%.230s" ... `len(SYSFS_CLASS_NET) + len("/%.230s/bridge") must be < SYSFS_PATH_MAX` or something like that.  But otherwise looks OK to me.

commodo:
makes sense 
will add

commodo:
funny story about this;
i was gonna apply here to package mstpd in Debian:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=767013

but then noticed i did not yet fix the GCC7 errs :p

PR #54, opened 2017-12-08
dump_msg: Avoid possible invalid pointer dereference in case of erronous message

simias:
A pointer was dereferenced before the message length was validated.

Very minor and extremely unlikely to happen in practice.


commodo:
[LGTM](https://lgtm.in/i/CplqZbVjq)

commodo:
thanks for the patch :)

PR #55, opened 2017-12-11
mstp.c: default to rstp proto level

commodo:
It's what most people seem to use.
And we do report it as being compliant with IXIA ANVL RSTP
test suite.

So, might as well default to RSTP and use mstpctl to
change to MSTP.


jonasj76:
Looks good according to me.

commodo:
since no objections, i'll merge this ;

PR #56, opened 2017-12-11
Correct the payload type of IFLA_OPERSTATE from int to __u8

simias:
Since the data is padded with zeroes to 4-byte boundaries it effectively worked on little endian machines, although unless I'm missing something it wouldn't work at all on big endian platforms.

https://github.com/torvalds/linux/blob/50c4c4e268a2d7a3e58ebb698ac74da0de40ae36/net/core/rtnetlink.c#L1555

I suspect it might also create an out-of bound access if this is the last entry in the buffer (since the payload is only one byte). This is rather unlikely however.

### DIFF
--- a/brmon.c
+++ b/brmon.c
@@ -105,7 +105,7 @@ static int dump_msg(const struct sockaddr_nl *who, struct nlmsghdr *n,
 
     if(tb[IFLA_OPERSTATE])
     {
-        int state = *(int*)RTA_DATA(tb[IFLA_OPERSTATE]);
+        __u8 state = *(__u8*)RTA_DATA(tb[IFLA_OPERSTATE]);
         switch (state)
         {
             case IF_OPER_UNKNOWN:


commodo:
@simias 
the cast definitely makes sense to be converted to `u8` since that's what the kernel type is for `IFLA_OPERSTATE`

this code did work well on PPC32 and PPC64 [both big-endian] last time i checked ;
but I do admit there is a chance for this to work incorrectly because the `(int *)` cast would cover more than 1 byte and the extra bytes could be garbage

thanks for the patch :)

[LGTM](http://25.media.tumblr.com/a77d4c6e492318a0010bf750eac2a101/tumblr_mjf4xt19vr1roja4ko1_500.gif)

simias:
Interesting, I guess it's been long enough since I've last worked on a BE machine that I don't remember how they work anymore...

commodo:
well, it could have been a fluke ;
in any case, over-casting [or casting and dereferencing a higher pointer] is a valid issue

i hope you don't mind the cat-gif ;
i try to use humor wherever i can, to avoid some day-to-day mundane-ness

simias:
Oh not at all, your repo your rules!

PR #57, opened 2017-12-12
set_br_up: fix info message to actually show the transition

simias:
Previously both the pre- and post-state variables were set to one
another before the message was printed.


simias:
Sorry for the many small PRs, I create them as I encounter something while I browse the code.

commodo:
no worries: keep'em coming :)
so far you're doing a great job

PR #61, opened 2018-03-08
Add STP filtering per port, courtsey of Cumulus Networks.

nihilus:
This adds the ability to turn off STP operation on a per-port basis.

### DIFF
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -132,6 +132,7 @@ typedef enum {
     PARAM_BPDUGUARDERROR,
     PARAM_NETWORKPORT,
     PARAM_BA_INCONSISTENT,
+    PARAM_BPDUFILTERPORT,
 } param_id_t;
 
 typedef struct {
@@ -498,6 +499,7 @@ static const cmd_param_t cist_port_params[] = {
     { PARAM_BPDUGUARDERROR, "bpdu-guard-error" },
     { PARAM_NETWORKPORT,    "network-port" },
     { PARAM_BA_INCONSISTENT,"ba-inconsistent" },
+    { PARAM_BPDUFILTERPORT, "bpdufilter-port" },
 };
 
 static int detail = 0;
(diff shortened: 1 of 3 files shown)


ctl_main.c
```diff
@@ -1164,6 +1170,17 @@ static int cmd_setportdonttxmt(int argc, char *const *argv)
     return set_port_cfg(dont_txmt, getyesno(argv[3], "yes", "no"));
 }
 
+static int cmd_setportbpdufilter(int argc, char *const *argv)
+{
+    int br_index = get_index(argv[1], "bridge");
+    if (0 > br_index)
+        return br_index;
+    int port_index = get_index(argv[2], "port");
```
commodo: maybe declare `int port_index;` at the beginning of the function;
this is for cases where some people may use some old compilers, or compiler options that don't enable C99

mstp.c
```diff
@@ -3254,9 +3270,10 @@ static bool PTSM_run(port_t *prt, bool dry_run)
                 return false;
             if(prt->sendRSTP)
             { /* implement MSTP */
-                if(prt->newInfo || (prt->newInfoMsti && !mstiMasterPort)
+                if(!prt->dontTxmtBpdu && !prt->bpduFilterPort &&
```
commodo: this also squeezes the `!prt->dontTxmtBpdu` condition in;
which didn't seem to be there before; and has no explanation why it's added

ctl_main.c
nihilus: Maybe it should be declared that way in the rest of the codebase then?

ctl_main.c
```diff
@@ -579,6 +581,7 @@ static int do_showport(int br_index, const char *bridge_name,
                 printf("Num RX TCN           %u\n", s.num_rx_tcn);
                 printf("  Num Transition FWD %-23u ", s.num_trans_fwd);
                 printf("Num Transition BLK   %u\n", s.num_trans_blk);
+                printf("  bpdufilter port    %-23s\n", BOOL_STR(s.bpdu_filter_port));
```
mg964: How about "BPDU Filter Port" for the display string?

commodo:
@nihilus 
Can you re-spin this patch ?

And I'll merge it

Thanks
Alex

nihilus:
Yes, I'll try to fix it asap

mg964:
I have an updated version of this patch - added a counter of dropped BPDUs. Want me to generate a new PR or I can post a few comments on this thread?

mg964:
AFAICT I don't have the necessary privs to perform a direct push into this PR. I've generated a new PR (#72):

https://github.com/mstpd/mstpd/pull/72

commodo:
Well, you should have been able to force-push on your own Github fork.
In case, let's close this as superseeded by #72

commodo:
Ah. I think I confused  @nihilus with @mg964 ; because both names have MG initials
Sorry about that.

PR #63, opened 2018-04-25
Close control socket on termination

mg964:
On issuing a /sbin/mstp_restart command the daemon would occasionally
fail with "error, server_socket: Couldn't bind socket: Address already
in use".

On termination explicitly close the server socket
(ctl_socket_cleanup()).


commodo:
merged
thanks

PR #64, opened 2018-04-26
Rearrange the startup sequence

mg964:
Rearrange the startup sequence so that the command line arguments are
processed first followed by the opening of the syslog connection. In
this way the initial self-check sequence (and version number) gets
pushed out to syslog.

In addition include the PID in syslog messages:

mstpd[21648]: main: Version - 0.0.6-
mstpd[21648]: main: Sanity checks succeeded

### DIFF
--- a/main.c
+++ b/main.c
@@ -87,33 +87,35 @@ int signal_init(void)
     return 0;
 }
 
+static int sanity_check(void)
+{
+    bridge_identifier_t BridgeIdentifier;
+    mst_configuration_identifier_t MST_ConfigurationIdentifier;
+
+    TST(sizeof(BridgeIdentifier) == 8, -1);
+    TST(sizeof(BridgeIdentifier.u) == 8, -1);
+    TST(sizeof(BridgeIdentifier.s) == 8, -1);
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -146,6 +123,35 @@ int main(int argc, char *argv[])
...
 
+    if(daemonize)
+        print_to_syslog = 1;
+
+    if(print_to_syslog)
+        openlog(APP_NAME, LOG_PID, LOG_DAEMON);
+
+    INFO("Version - " PACKAGE_VERSION "\n");
+
+    /* Sanity check */
```
commodo: since you're moving this, could you make the sanity checks into a function and call it here ?
it would be nice to unclutter things a bit here

commodo:
otherwise looks good

main.c
mg964: Sure

commodo:
just as an update/heads-up: I will merge this even though the Travis CI build is failing ;
i fixed it in PR #65 [ which will be merged tomorrow ]
the CI has a way of moving itself under our feet [every now-n-then], which is not bad, since it means we're forced to update stuff

PR #65, opened 2018-04-26
.travis.yml: switch to Ubuntu Xenial distro for testing

commodo:
Not sure which is the default distro in Travis CI at the moment.
But, it allows the use of Ubuntu Xenial (16.04 which is great because
that's pretty recent compared to Trusty (14.04) or Precise (12.04)


commodo:
if no objections, will merge this tomorrow

PR #66, opened 2018-07-02
Check STP state before terminating MSTPd

mg964:
When STP is turned off on an interface, check all other bridge/switch
instances to see if user-mode STP is enabled. Only terminate the
daemon (SIGTERM) if there are no other bridge/switch interfaces in use
by the daemon.

### DIFF
--- a/bridge-stp.in
+++ b/bridge-stp.in
@@ -185,6 +185,7 @@ case "$action" in
             fi
 
             # If bridge is in user_stp mode, then it is probably using MSTP.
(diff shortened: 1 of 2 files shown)


PaulSD:
:+1: LGTM

commodo:
the change is fine
seems shellcheck started complaining about something new;

will update that in a separate change

thanks :)

mg964:
Yeah saw the complaints, couldn't make sense of them.

PR #70, opened 2018-07-06
ctl_main: Add JSON output format support

namedun:
Added JSON output format for all available show commands:
  - showbridge
  - showmstilist
  - showmstconfid
  - showvid2fid
  - showfid2mstid
  - showport
  - showportdetail
  - showtree
  - showtreeport
  - showmstconfid

This feature is very useful when we need obtain parsed data from mstpd.
For example, we can obtain JSON data in CGI script and transfer it
to the client's web browser for further analysis or/and display.

Output format can be selected by -f <format> (--format=<format>) command
line option. Currently available formats:
  - plain : plain text (used by default)
  - json  : JSON

Signed-off-by: Anton Kikin <a.kikin@tano-systems.com>

### DIFF
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -67,6 +67,33 @@ static inline int get_id(const char *str, const char *doc, unsigned int max_id)
     return id;
 }
 
+/* Supported output formats */
+typedef enum {
+    FORMAT_PLAIN, /* plain text (default) */
+    FORMAT_JSON   /* JSON */
+} format_id_t;
+
+/* Default output format */
+static format_id_t format = FORMAT_PLAIN;
+
+static inline void do_arraystart_fmt(void)
+{
+    if (FORMAT_JSON == format)
+        printf("[");
+}
+
+static inline void do_arrayend_fmt(void)
+{
+    if (FORMAT_JSON == format)
+        printf("]");
+}
+
+static inline void do_arraynext_fmt(void)
+{
+    if (FORMAT_JSON == format)
+        printf(",");
+}
+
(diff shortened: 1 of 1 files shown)


commodo:
Overall looks very good.
You could squash the 2 Codacy commits into a single one; since they're not merged yet, you could do.

I'll try to make some time to run this a bit.
Maybe also adds some auto-tests for this.

Ping me next week if I become silent.

Thank you very much for this work :)
Looks awesome.

mg964:
Was any consideration given to using a formal JSON library (json-c) in order to produce the output?

commodo:
I also thought about something like json-c being used.
But if it's already made the effort to do it without it, it's fine as is.
One less dependency to care about ; or, we don't need strict dependency on external libs.

We could think about adding it as an optional dependency at some point [if it makes sense]; like --with-libjson-c or something like that.

I tried it out now and this code seems to run nicely.
It doens't have JSON pretty-print, but you can pipe stuff via `python -mjson.tool` if you need pretty printing.
So,  `mstpctl -f json showbridge/showport | python -mjson.tool` works neatly.

I'll admit, it's an external dependency [i.e. python] for pretty-printing ; but we already need Python for `ifquery`

commodo:
@mg964 
I don't want to override you or anything; if you have open-comments, we can discuss them :)

Thanks for prompting the message today; it prompted me to look into this PR today.

mg964:
No that's fine, it was as much a curiosity as anything else. Eliminating the dependency is fine.

PR #72, opened 2018-11-14
Disable STP processing on a specific port

mg964:
Add support for a "BPDU Port Filter" option ("setportbpdufilter"),
i.e. discard BPDUs on ingress and do not issue BPDUs. Further add a
counter that indicates the number of received BPDUs that have been
discarded.

### DIFF
--- a/mstp.c
+++ b/mstp.c
@@ -129,6 +129,7 @@ static void port_default_internal_vars(port_t *prt)
     assign(prt->rapidAgeingWhile, 0u);
     assign(prt->brAssuRcvdInfoWhile, 0u);
     prt->BaInconsistent = false;
+    prt->num_rx_bpdu_filtered = 0;
     prt->num_rx_bpdu = 0;
     prt->num_rx_tcn = 0;
     prt->num_tx_bpdu = 0;
@@ -300,6 +301,7 @@ bool MSTP_IN_port_create_and_add_tail(port_t *prt, __u16 portno)
     prt->BpduGuardError = false;
     prt->NetworkPort = false;
     prt->dontTxmtBpdu = false;
+    prt->bpduFilterPort = false;
     prt->deleted = false;
 
     port_default_internal_vars(prt);
@@ -500,6 +502,7 @@ void MSTP_IN_set_port_enable(port_t *prt, bool up, int speed, int duplex)
             prt->portEnabled = true;
             prt->BpduGuardError = false;
             prt->BaInconsistent = false;
+            prt->num_rx_bpdu_filtered = 0;
             prt->num_rx_bpdu = 0;
             prt->num_rx_tcn = 0;
             prt->num_tx_bpdu = 0;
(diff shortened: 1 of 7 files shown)


nihilus:
Actually counters is unnecessary overhead since this is being enforced by the user, and eventually they will wrap.

mstp.c
```diff
@@ -3252,6 +3275,10 @@ static bool PTSM_run(port_t *prt, bool dry_run)
             }
             if(!(prt->txCount < prt->bridge->Transmit_Hold_Count))
                 return false;
+
+            if(prt->bpduFilterPort)
+                return false;
```
commodo: as a note, this patch here, differs a bit from the initial one, which was here: https://github.com/mstpd/mstpd/pull/61/commits/83eaaf81d74c954301a1df62429556d44be135d1#diff-afc426ea39d24015ac2bbcb1eed6a712L3256

mg964:
"unnecessary overhead"? For me "bpdu-filter" is a bit of additional paranoia (along with bpdu-guard), in a well behaved network you would expect the counter to remain at 0. But if the counter is incrementing you may want to take further action - "why has this edge node started sending me BPDUs".

mstp.c
mg964: Yeah, seemed prudent to perform the check before the send RSTP/STP check. Can move the check if desired.

commodo: No, that's fine.
Your patch makes a bit more sense than the initial one.

commodo: i.e. I understand it easier

commodo:
The BPDU filter counters are a neat addition.
So, let's have them ; they can be useful when debugging the state-machine of STP.

commodo:
[LGTM](https://i.pinimg.com/564x/80/dc/fe/80dcfe033af4ceea0c2611f3c9375b65.jpg)

commodo:
Will merge this later today if no objections :)

commodo:
Hmm, let's move this to tomorrow ; it gives other timezones a chance to look.
[ Sorry for the rush ]

commodo:
@nihilus - are you referring to disabling just BPDU filter counters via configure-option ?

regarding "individual interfaces per port specific for STP (weird implementation)." - we also used something similar ; it's a pretty good way to use a switch ASIC to do switching and control bridging protocols (STP, LLDP) per-port with lldpd & mstpd ;

nihilus:
@commodo Yes, exactly :-)

commodo:
Well, I'm just trying to clarify what @nihilus is saying.

Personally, I'm not convinced that a `disabling just BPDU filter counters via configure-option` is a good idea. It's a bit like trying to use a sledge-hammer to kill a fly.
The overhead that they add is very minor, and a config option is the proverbial sledge-hammer.

I'd say, let's have this change as-is for now.
Maybe we can find a clearer reason later if we need something like this.

PR #74, opened 2018-12-27
fix to build with musl library

fidomax:


commodo:
Many thanks for your patch :)

When I was working at my previous job, we were building this with musl,since we were building on top of OpenWrt.
Now, we could probably try to integrate a build using the OpenWrt SDK [which should have musl].

But I keep wanting to officially package mstpd for OpenWrt at some point in time :)

Anyway, I usually try to also add some humor when approving code.
[LGTM](https://i.pinimg.com/564x/ad/9c/5d/ad9c5d36eedc886d43f82596d8e4e087.jpg)

PR #75, opened 2019-01-10
fread() return value checking fixed

ZolotarevIgor:
https://en.cppreference.com/w/c/io/fread


commodo:
I don't think that function (`rtnl_from_file()`) is used anywhere.
So, I'm not sure if it's worth patching anyway.
That function arrived there because libnetlink was probably copied from an upstream library and reduced in size.

Your patch looks good, but we can't test it [since it's unused code].

If you want, we can remove the `rtnl_from_file()` function altogether.
I'd be fine with that; it causes less confusion when looking through the code.

ZolotarevIgor:
Ok, I get it, thank you.

PR #76, opened 2019-01-23
get_id: id overflow fixed

ZolotarevIgor:
id values in range [LONG_MIN + 1; LONG_MIN + max_id] were converted to valid id. 
Then:
zolotarev@localhost: ~ / mstpd> ./mstpctl createtree br0 -- -9223372036854771714
2019-01-23 16:29:48 MSTP_OUT_set_state: br0:enp7s0:4094 entering blocking state
2019-01-23 16:29:48 MSTP_OUT_flush_all_fids: br0:enp7s0:4094 Flushing forwarding database
Now:
zolotarev@localhost: ~ / mstpd> ./mstpctl createtree br0 -- -9223372036854771714
Bad mstid -9223372036854771714

### DIFF
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -56,10 +56,8 @@ static inline int get_index(const char *ifname, const char *doc)
 
 static inline int get_id(const char *str, const char *doc, unsigned int max_id)
 {
-    int id = strtol(str, NULL, 10);
-    if((0 > id) || (max_id < id)
-       || ((0 == id) && ('0' != str[0]))
-      )
+    unsigned long id = strtoul(str, NULL, 10);
+    if((max_id < id) || ((0 == id) && ('0' != str[0])))
     {
         fprintf(stderr, "Bad %s %s\n", doc, str);
         return -1;


ctl_main.c
```diff
@@ -56,10 +56,8 @@ static inline int get_index(const char *ifname, const char *doc)
 
 static inline int get_id(const char *str, const char *doc, unsigned int max_id)
 {
-    int id = strtol(str, NULL, 10);
-    if((0 > id) || (max_id < id)
-       || ((0 == id) && ('0' != str[0]))
-      )
+    unsigned long id = strtoul(str, NULL, 10);
```
commodo: i'm not sure that fixing this is worth the effort much;
i admit the existing code is not perfect with respect to extreme/corner cases, but if we really want to fix or address this, then a bit more work is required;

we'd probably need some function that returns true/false if the string is completely numerical, and has a length of maximum 4 characters; that way you can avoid other corner cases where you do `1234abcd` as test-input, or `abcd1234`, or `123` or `-123`, etc

if i'm reading through various docs for `strtoul` things differ slightly between what's mentioned in the return value sections;
see  http://www.cplusplus.com/reference/cstdlib/strtoul/ &   http://man7.org/linux/man-pages/man3/strtoul.3.html

the section about the leading minus sign is un-mentioned in one of the docs, and you also need to take into account that there are several others libc implementations [uClibc, musl-libc, glibc, etc ]
i usually like to give out this link http://www.etalabs.net/compare_libcs.html  [because i like musl]

in any case, i am not against fixing this, but we should go all the way if we're doing it, because this change is touching on a corner case and probably omitting other corner cases :)

ZolotarevIgor: OK, thank you, I'll think about all of this

commodo:
Let's close this for now.
Feel free to open a new PR with changes.

Thanks
Alex

PR #77, opened 2019-06-03
mstpctl: implement batch command processing

### LINKED ISSUE #4: Add batch command to mstpctl
There is a need to add some command which enables batch parameters setting. It can be command-line option which gives the name of the file containing list of ordinary commands.
Main goal: implement many settings in one batch. E.g. when assigning VIDs to FIDs it is too long to assign it one-by-one. We already have internal interface to set the whole mapping table in one bulk, there is only lack of user interface for the task.

commodo:
Fixes #4 

Changeset implements batch command processing.
This can be specified via stdin (with -s / --stdin CLI specifier) or from file (with -b / --batch <file> ).

By default processing is done until the first failing or invalid command is found.
A  -i / --ignore option can be provided to skip faulty commands ; this can be useful in teardown cases.

For batch-files, they are validated first to check if there are any invalid commands/typos, but these will be ignored if `--ignore` is provided.
For stdin, all batch commands willl be processed until the first failing; depending on whether `--ignore` is provided, processing will stop or continue.

### DIFF
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -2301,6 +2301,9 @@ static void help(void)
     printf("  -h | --help              Show this help text\n");
     printf("  -V | --version           Show version\n");
     printf("  -b | --batch <file>      Process file with mstpctl commands\n");
+    printf("  -s | --stdin             Process mstpctl commands from stdin\n");
+    printf("                           Make sure to provide newlines between\n");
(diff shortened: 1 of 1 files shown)


commodo:
Changelog v1 -> v2:
* fix codacy warning

commodo:
Will merge this later this week if no objections

PR #82, opened 2019-10-10
treewide: allow room for null-char in strncpy()

commodo:
strncpy() is not safe as it can put `len` elements of a string and discard
the null-char if there isn't any room left.

An alternative option would be to use `strlcpy()` but that function may not
be portable to all libc implementations.
So, just reduce the `len` by 1 byte to allow room for null-char.

This was reported via string-fortify headers.


mg964:
> Well, we could, but who knows? I'm not sure.
> I could not find any reference about it's portability, and some references said it's not specified via ANSI C, so I did not bother to use it.

Ack. 

There is this: https://lwn.net/Articles/612244/, the implication being that strlcpy() was going to be added to glibc, but I can't actually see any evidence (https://sourceware.org/glibc/wiki/FAQ#Why_no_strlcpy_.2BAC8_strlcat.3F).

PR #90, opened 2020-06-03
debian: remove packaging rules and add initial readme

commodo:
This change removes the packaging rules for debian.
They will continue to live in 'debian/xxxx' branches from now on.
Also, removing with this change is the travis-ci rules to do any debian
build checks in lintian.


commodo:
@vincentbernat  ping
this is a simple PR;
but mostly to notify you and get things started;

i had a bit of fear-freeze where i tried to figure more things before getting the PR going;
now, just removing from master; and adding a readme;

will continue from `debian/sid` branch
let's see if i missed anything in the CI part

PR #91, opened 2020-06-03
debian/sid: update packaging info

commodo:
Some baby-steps [from my side] to start with some Debian packaging updates.

Pinging @vincentbernat for review/help.
[ We may have to add you as a member to the group for this review stuff ]

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,13 @@
 mstpd (0.0.9) unstable; urgency=medium
 
+  [ Alexandru Ardelean ]
+  * d/control: bump Standards-Version.
+
+  [ Eemil Lagerspetz ]
   * Handle increasing/decreasing fdelay and maxage in sync in ifcfg files.
 
  -- Eemil Lagerspetz  <eemil.lagerspetz@lempea.com>  Thu, 28 May 2020 14:18:10 +0300
+
 mstpd (0.0.8) unstable; urgency=medium
 
   * Build Debian package.
--- a/debian/changelog
+++ b/debian/changelog
@@ -2,6 +2,7 @@ mstpd (0.0.9) unstable; urgency=medium
 
   [ Alexandru Ardelean ]
   * d/control: bump Standards-Version.
+  * Set upstream metadata fields: Repository, Repository-Browse.
 
   [ Eemil Lagerspetz ]
   * Handle increasing/decreasing fdelay and maxage in sync in ifcfg files.
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Section: net
 Priority: optional
 Maintainer: Vitalii Demianets <vitalii@orsoc.se>
 Build-Depends: debhelper (>=9), dh-autoreconf
-Standards-Version: 3.9.8
+Standards-Version: 4.1.1.0
 Homepage: https://github.com/mstpd/mstpd
 Vcs-Git: https://github.com/mstpd/mstpd.git
 Vcs-Browser: https://github.com/mstpd/mstpd
--- a/debian/upstream/metadata
+++ b/debian/upstream/metadata
@@ -1,0 +1,2 @@
+Repository: https://github.com/mstpd/mstpd
+Repository-Browse: https://github.com/mstpd/mstpd


vincentbernat:
OK, for the branch to work with Debian tools, it needs to "rooted" to a tag. Let's keep it as a PR for the moment, but when the moment is right, maybe you can do a release 0.0.9 of mstpd and we will move debian/sid to this release (either by rebasing if it's still a PR, or by merging 0.0.9).

- d/changelog: version should be `0.0.9-1` (`-1` is Debian local version)
- d/changelog: only keep one entry for the initial release (into Debian) with `Initial release. Closes: #767013.`
- d/compat: remove it
- d/control: build-depends on `debhelper-compat (= 12)` instead of `debhelper`
- d/control: you cannot depend on python, you can depend on python3 if needed (but is that needed?)
- d/control: `Vcs-Browser` should point to https://github.com/mstpd/mstpd/tree/debian/sid while `Vcs-Git` should get `-b debian/sid`.
- d/control: the "Maintainer" will get all the emails about bugs and stuff like that. This could be changed to a team email, like `team+debian-mstpd-team AT tracker.debian.org` that can be managed through tracker.debian.org.
- d/copyright: it seems all good!
- d/prerm: it should be removed, you cannot kill process by names, I would suggest having a systemd unit and let bridge-stp start mstpd this way
- d/rules: remove all the boilerplate comments, remove the hardening flags (not needed anymore).

As for `/lib` ignore my last comment, this is the correct place. `/usr/share` is for data.

Also, is the long description still true with the support of VLAN-aware bridges?

commodo:
thanks for the feedback-list;
i'll try to allocate some time for it, in the next few days;

in the meantime, some replies;
if i missed anything, it is by accident; nothing indented to avoid any replies;

> * d/compat: remove it
> 
> * d/control: build-depends on `debhelper-compat (= 12)` instead of `debhelper`

unless @PaulSD has something to object here, i don't have a problem with this;
maybe as a mid-way resolution we could do another branch with the compat level for debhelper 9

>     * d/control: the "Maintainer" will get all the emails about bugs and stuff like that. This could be changed to a team email, like `team+debian-mstpd-team AT tracker.debian.org` that can be managed through tracker.debian.org.

is this an example, or an actual working email that is somehow working via some Debian infrastructure?
would a list of emails work?  email1, email2, etc?



>     * d/prerm: it should be removed, you cannot kill process by names, I would suggest having a systemd unit and let bridge-stp start mstpd this way

i was trying to see about how to do this; with taking example exabgp
there's an exabgp.init & exabgp.install that seems to be related;
if you have a simple [for beginners] example; it's appreciated;



> Also, is the long description still true with the support of VLAN-aware bridges?

if you're referring to this https://github.com/mstpd/mstpd/blob/master/README.VLANs.md
then yes;
@PaulSD wrote that and documented it; but not much has changed [to invalidate it] since then, so i would stay it is still valid;

vincentbernat:
❦  3 juin 2020 22:49 -07, Alexandru Ardelean:

>> * d/control: build-depends on `debhelper-compat (= 12)` instead of `debhelper`
>
> unless @PaulSD has something to object here, i don't have a problem with this;
> maybe as a mid-way resolution we could do another branch with the
> compat level for debhelper 9

If you want to support older versions of Debian/Ubuntu, yes, this can be
done in another branch.

>>     * d/control: the "Maintainer" will get all the emails about bugs
>> and stuff like that. This could be changed to a team email, like
>> `team+debian-mstpd-team AT tracker.debian.org` that can be managed
>> through tracker.debian.org.
>
> is this an example, or an actual working email that is somehow working
> via some Debian infrastructure?

Yes.

> would a list of emails work?  email1, email2, etc?

Alternatively, you can put one email in Maintainer and a list of emails
in Uploaders.

>>     * d/prerm: it should be removed, you cannot kill process by names, I would suggest having a systemd unit and let bridge-stp start mstpd this way
>
> i was trying to see about how to do this; with taking example exabgp
> there's an exabgp.init & exabgp.install that seems to be related;
> if you have a simple [for beginners] example; it's appreciated;

I'll do a PR for this. Do you also want to support SysV init?

commodo:
> I'll do a PR for this. Do you also want to support SysV init?

umm; SysV is a good question; tbh, i don't know about the users of this package to say for sure;
maybe let's not add it for now, and we can probably add it later [if we get a request];
i'm trying to think of this in a way for going forward;
since this hasn't been packaged into Debian [officially], maybe it shouldn't start adding legacy stuff;

also, this package is a bit tricky [on it's own], since the kernel calls `/sbin/bridge-stp` which does some configuration with the daemon; which we need to be careful of, because it can have some race-conditions;

commodo:
I'll be on vacation next week.
So, likely I won't be able to answer anything here.

Sorry about that.
The vacation was planned a few weeks back.
Will ping when back.
I have another vacation, which is more of an office shutdown in the last week of June.
[some companies like to do it as a means to save money]
Let's see how this goes.

commodo:
@vincentbernat 
sorry for being late on this;
it's been a while now since this was started;

anything needed on this PR to move forward?

mg964:
> d/control: you cannot depend on python, you can depend on python3 if needed (but is that needed?)

Since support for python2 is being dropped in debian 11, any reason not to simply move to python3? The only reference is utils/ifquery (and control):

```
#!/usr/bin/python
# Crudely emulate ifquery on systems that don't have it.
```

The big question is, do we still need to provide a substitute ifquery?

commodo:
> > d/control: you cannot depend on python, you can depend on python3 if needed (but is that needed?)
> 
> Since support for python2 is being dropped in debian 11, any reason not to simply move to python3? The only reference is utils/ifquery (and control):
> 
> ```
> #!/usr/bin/python
> # Crudely emulate ifquery on systems that don't have it.
> ```
> 
> The big question is, do we still need to provide a substitute ifquery?

i was thinking about replacing ifquery with a shell implementation;
just never got around to it;

python is nice, but it's a bit dependency; on cloud VMs it's not a big deal [you get Python for free];
on embedded systems, it's large;

pavlix:
> # Crudely emulate ifquery on systems that don't have it.

Which are the Debian based systems (with `/etc/network/interfaces`) that don't have it? Cannot we just drop it?

commodo:
> Which are the Debian based systems (with `/etc/network/interfaces`) that don't have it? Cannot we just drop it?

i think ifquery was written based on Cumulus Networks's OS, which is Debian based;
maybe moving forward we could remove it;

we can start a PR as a discussion;
personally, i've never used it;
because i was running mstpd on OpenWrt and was using some netifd + other logic;

PaulSD:
The `ifquery` script is used on RPM-based distros.  It is not needed on Debian based distros.

PR #94, opened 2020-06-11
ifquery: Use versioned python shebang and fix flake8 warnings

PaulSD:
`utils/ifquery` is used by rpm.spec for mstpd configuration on RPM-based systems.  `rpmbuild` now requires all python shebangs to be versioned ("python2" or "python3").

While I'm changing things in that file, the `flake8` python linter complains about some minor formatting issues, so I've cleaned those up.

### DIFF
--- a/utils/ifquery
+++ b/utils/ifquery
@@ -4,29 +4,31 @@
 import sys
 import glob
 
+
 def parse_config(interfaces_file, interface):
     f = open(interfaces_file)
     intf = ''
     for line in f:
         str = line.strip().split(' ')
         len_str = len(str)
-        if 2 > len_str :
+        if 2 > len_str:
             continue
-        if str[0].startswith('#') :
+        if str[0].startswith('#'):
             continue
         # Check if the source <file> keyword is present and extract the file
-        if str[0] == 'source' :
+        if str[0] == 'source':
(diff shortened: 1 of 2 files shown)


utils/ifquery
```diff
@@ -1,32 +1,34 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
commodo, line 1: i'd argue we'd get [a few] complaints for a while from users wanting to run `master` on older systems;
i'm not fully aware what's the cutoff from 2 to 3 on all distros;
maybe we should let the default be `python` and let distros decide when `python` points to `python3`

PaulSD: As far as I'm aware, this script is only being used on RPM-based distros (I wrote it specifically for use in RPMs).  And as far as I'm aware, I'm the only one actually building RPMs for `mstpd` at this time.
But if you're worried about it, I suppose I could have the `rpm.spec` file edit the `ifquery` file during the build.

commodo: > As far as I'm aware, this script is only being used on RPM-based distros (I wrote it specifically for use in RPMs).

a bit hard to say whether you are the only one :)
the feedback we get is silence when things work;
i'm not against switching the shebang to `python3`;
we can do that;
i'll leave the preference up to you;

PR #101, opened 2020-10-04
Copy the system unit build artifact from the build folder

veedo:
The final service file is output into the build directory, not the source directory


commodo (approved):
good catch

commodo:
thanks :)

PR #103, opened 2021-01-22
RFC: add a driver setting MST per port state as linux per vlan port state

KanjiMonster:
Since release 5.6 Linux supports a per VLAN STP state in addition to the bridge port stp state.

This driver attempts to make use of it to propagate the MSTI port states to their appropriate VLAN STP states using the new netlink messages for this.

Our understanding of MSTP is currently very vague, and we are trying to understand how this work and whether MSTPd does the right thing, and if this implementation is enough, or more work is needed. 

Some notes for the implementation:

* I imported two commits for libnetlink from iproute2, where it seems to have originated. I kept all authorship intact, and added my SoB since I slightly modified them so they apply.
* We verified that setting the per VLAN STP state does work, but of course checking isn't easy since there are no userspace utilities yet to show the per VLAN STP states. We have an internal patched iproute2 bridge utility for that, but it's far from an upstreamable state yet.
* I haven't found an easy way to find out the VLANs of a MSTI, so for now I just iterate over all possible VIDs and check if VID -> FID -> MSTID matches the current MSTI. Not efficient, but at least a fixed runtime.
* While it is possible to set the STP state of VLAN ranges and multiple VLANs at once, since the kernel will discard the message on the first invalid VLAN of a port, sending individual messages is (probably) safer.
* AFAICT the way the bridge port STP state and the VLAN STP state interact is that the "lower" of the two wins. E.g. only if both are in forwarding, the vlan for the port will forward. So setting the bridge port STP state to blocked will block for all VLANs, regardless of their per VLAN STP states.

Now to the RFC part:

Currently MSTPd sets the CIST state as the bridge port STP state. This e.g. results in all VLANs being blocked if the CIST says the port is blocked, even if other MSTIs say the port should forward for those VLANs.

Does this make sense? Can this even happen? Or should the CIST state only being applied to the VLANs not assigned to any MSTIs? If the latter, then MSTPd would need to be changed to not set the CIST state as the bridge port stp state.

Is it correct that MSTP(d) expects a VLAN to be present/available on all bridge interfaces? If so, then we need to make MSTPd listen for VLANDB entries and dynamically apply STP state if they get created, else STP state might get lost if VLANs get created after MSTPd applied the VLAN state of a MSTI.

And more a understanding question of MSTP, since AFAICT all bridges in a MST are treated as one bridge for the CIST, what does that mean for the CIST port states within the region? Are they all forwarding, or should they be also be set to some non-looping state?

Looking at figure 13-11 of 802.11Q-2018 (p. 496), there on the CIST the port of bridge 84 to LAN J (to bridge 97) is blocked, but in the MSTI topology it is forwarding.

If this means that the MTSI should be in forwarding, then this answers the first question as in MSTPd should not set the CIST state to the bridge port STP state, but the VLAN STP states that are part of the CIST / not assigned to a MSTI.

### DIFF
--- a/linux_pvstate/driver_deps.c
+++ b/linux_pvstate/driver_deps.c
@@ -1,0 +1,167 @@
+/*
+ * driver_deps.c    Propagate MSTI port states to Linux Per VLAN STP states
+ *
+ *  This program is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU General Public License
+ *  as published by the Free Software Foundation; either version
+ *  2 of the License, or (at your option) any later version.
+ *
+ * Authors: Vitalii Demianets <dvitasgs@gmail.com>
+ * Authors: Jonas Gorski <jonas.gorski@bisdn.de> -- Set Linux Per VLAN state
+ */
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <string.h>
+#include <asm/byteorder.h>
+
+#include <linux/if_bridge.h>
+
+#include "bridge_ctl.h"
+#include "libnetlink.h"
+#include "log.h"
+#include "mstp.h"
+
+static int br_set_vlan_state(struct rtnl_handle *rth, unsigned ifindex, __u16 vid, __u8 state)
+{
+    struct
+    {
+        struct nlmsghdr n;
+        struct br_vlan_msg bvm;
+        char buf[256];
+    } req;
+    char entry_buf[256];
+    struct rtattr *rta = (void *)entry_buf;
+    struct bridge_vlan_info vlan_info;
+    struct rtattr *nest;
+
+    memset(&req, 0, sizeof(req));
+
+    req.n.nlmsg_len = NLMSG_LENGTH(sizeof(struct br_vlan_msg));
+    req.n.nlmsg_flags = NLM_F_REQUEST | NLM_F_REPLACE;
+    req.n.nlmsg_type = RTM_NEWVLAN;
+    req.bvm.family = AF_BRIDGE;
+    req.bvm.ifindex = ifindex;
+
+    rta->rta_type = BRIDGE_VLANDB_ENTRY;
+    rta->rta_len = RTA_LENGTH(0);
+
+    vlan_info.vid = vid;
+    vlan_info.flags = BRIDGE_VLAN_INFO_ONLY_OPTS;
+
+    nest = rta_nest(rta, sizeof(entry_buf), BRIDGE_VLANDB_ENTRY);
+    rta_addattr_l(rta, sizeof(entry_buf), BRIDGE_VLANDB_ENTRY_INFO, &vlan_info, sizeof(vlan_info));
+    rta_addattr8(rta, sizeof(entry_buf), BRIDGE_VLANDB_ENTRY_STATE, state);
+
+    rta_nest_end(rta, nest);
+
+    addraw_l(&req.n, sizeof(req.buf), RTA_DATA(rta), RTA_PAYLOAD(rta));
+
+    return rtnl_talk(rth, &req.n, 0, 0, NULL, NULL, NULL);
+}
+
+
+/* Initialize driver objects & states */
+int driver_mstp_init()
+{
+    return 0;
+}
+
+/* Cleanup driver objects & states */
+void driver_mstp_fini()
+{
+
+}
+
+/* Driver hook that is called before a bridge is created */
+bool driver_create_bridge(bridge_t *br, __u8 *macaddr)
+{
+    return true;
+}
+
+/* Driver hook that is called before a port is created */
+bool driver_create_port(port_t *prt, __u16 portno)
+{
+    return true;
+}
+
+/* Driver hook that is called when a bridge is deleted */
+void driver_delete_bridge(bridge_t *br)
+{
+
+}
+
+/* Driver hook that is called when a port is deleted */
+void driver_delete_port(port_t *prt)
+{
+
+}
+
+
+/*
+ * Set new state (BR_STATE_xxx) for the given port and MSTI.
+ * Return new actual state (BR_STATE_xxx) from driver.
+ */
+int driver_set_new_state(per_tree_port_t *ptp, int new_state)
+{
+    port_t *prt = ptp->port;
+    bridge_t *br = prt->bridge;
+    int i;
+
+    /* CIST will already be handled by common code */
+    if(0 == ptp->MSTID)
+        return new_state;
+
+    /* There is no MSTID -> VID mapping, so for now check all possible VIDs.
+     * This is probably not the most efficient, but at least a constant, and
+     * checking 4k should not take that long.
+     *
+     * The mapping is VID -> FID -> MSTID, so let's go that way.
+     */
+    for (i = 1; i <= MAX_VID; i++)
+    {
+        __u16 fid = br->vid2fid[i];
+
+        if (br->fid2mstid[fid] != ptp->MSTID)
+            continue;
+
+        if (0 > br_set_vlan_state(&rth_state, prt->sysdeps.if_index, i, new_state))
+            ERROR_MSTINAME(br, prt, ptp, "Couldn't set kernel if %i vid %i bridge state %i",
+                          prt->sysdeps.if_index, i, new_state);
+    }
+
+    return new_state;
+}
+
+bool driver_create_msti(bridge_t *br, __u16 mstid)
+{
+    /* TODO: send "create msti" command to driver */
+    return true;
+}
+
+bool driver_delete_msti(bridge_t *br, __u16 mstid)
+{
+    /* TODO: send "delete msti" command to driver */
+    return true;
+}
+
+void driver_flush_all_fids(per_tree_port_t *ptp)
+{
+    /* TODO: do real flushing.
+     * Make it asynchronous, with completion function calling
+     * MSTP_IN_all_fids_flushed(ptp)
+     */
+    MSTP_IN_all_fids_flushed(ptp);
+}
+
+/*
+ * Set new ageing time (in seconds) for the port.
+ * Return new actual ageing time from driver (the ageing timer granularity
+ *  in the hardware can be more than 1 sec)
+ */
+unsigned int driver_set_ageing_time(port_t *prt, unsigned int ageingTime)
+{
+    /* TODO: do set new ageing time */
+    return ageingTime;
+}
--- a/libnetlink.c
+++ b/libnetlink.c
@@ -611,6 +611,37 @@ int rta_addattr_l(struct rtattr *rta, int maxlen, int type,
 	return 0;
 }
 
+int rta_addattr8(struct rtattr *rta, int maxlen, int type, __u8 data)
+{
+	return rta_addattr_l(rta, maxlen, type, &data, sizeof(__u8));
+}
+
+int rta_addattr16(struct rtattr *rta, int maxlen, int type, __u16 data)
+{
+	return rta_addattr_l(rta, maxlen, type, &data, sizeof(__u16));
+}
+
+int rta_addattr64(struct rtattr *rta, int maxlen, int type, __u64 data)
+{
+	return rta_addattr_l(rta, maxlen, type, &data, sizeof(__u64));
+}
+
+struct rtattr *rta_nest(struct rtattr *rta, int maxlen, int type)
+{
+	struct rtattr *nest = RTA_TAIL(rta);
+
+	rta_addattr_l(rta, maxlen, type, NULL, 0);
+
+	return nest;
+}
+
+int rta_nest_end(struct rtattr *rta, struct rtattr *nest)
+{
+	nest->rta_len = (void *)RTA_TAIL(rta) - (void *)nest;
+
+	return rta->rta_len;
+}
+
 int parse_rtattr(struct rtattr *tb[], int max, struct rtattr *rta, int len)
 {
 	memset(tb, 0, sizeof(struct rtattr *) * (max + 1));
--- a/libnetlink.c
+++ b/libnetlink.c
@@ -631,6 +631,7 @@ struct rtattr *rta_nest(struct rtattr *rta, int maxlen, int type)
 	struct rtattr *nest = RTA_TAIL(rta);
 
 	rta_addattr_l(rta, maxlen, type, NULL, 0);
+	nest->rta_type |= NLA_F_NESTED;
 
 	return nest;
 }
--- a/libnetlink.h
+++ b/libnetlink.h
@@ -35,13 +35,23 @@ int addattr32(struct nlmsghdr *n, int maxlen, int type, __u32 data);
 int addattr_l(struct nlmsghdr *n, int maxlen, int type, const void *data,
               int alen);
 int addraw_l(struct nlmsghdr *n, int maxlen, const void *data, int len);
+int rta_addattr8(struct rtattr *rta, int maxlen, int type, __u8 data);
+int rta_addattr16(struct rtattr *rta, int maxlen, int type, __u16 data);
 int rta_addattr32(struct rtattr *rta, int maxlen, int type, __u32 data);
+int rta_addattr64(struct rtattr *rta, int maxlen, int type, __u64 data);
 int rta_addattr_l(struct rtattr *rta, int maxlen, int type,
                          const void *data, int alen);
 
 int parse_rtattr(struct rtattr *tb[], int max, struct rtattr *rta, int len);
 int parse_rtattr_byindex(struct rtattr *tb[], int max, struct rtattr *rta,
                          int len);
+
+struct rtattr *rta_nest(struct rtattr *rta, int maxlen, int type);
+int rta_nest_end(struct rtattr *rta, struct rtattr *nest);
+
+#define RTA_TAIL(rta) \
+		((struct rtattr *) (((void *) (rta)) + \
+				    RTA_ALIGN((rta)->rta_len)))
 
 #define parse_rtattr_nested(tb, max, rta) \
     (parse_rtattr((tb), (max), RTA_DATA(rta), RTA_PAYLOAD(rta)))


linux_pvstate/driver_deps.c
```diff
@@ -0,0 +1,167 @@
...
+#include <asm/byteorder.h>
+
+#include <linux/if_bridge.h>
+
+#include "bridge_ctl.h"
+#include "libnetlink.h"
+#include "log.h"
+#include "mstp.h"
+
+static int br_set_vlan_state(struct rtnl_handle *rth, unsigned ifindex, __u16 vid, __u8 state)
```
commodo, line 26: i'd merge this per-VLAN state code into the main driver_deps code;
maybe guarded via a global flag here;

`driver_set_per_vlan_state(bool enabled) `   or something like that

otherwise looks good;

commodo:
So, no-one else has replied.

From my side: I don't know how to answer the questions without digging deeper into the IEEE spec and reviewing the code.
It's been a while since I used mstpd, and now I just stick around to offer a presence.
Also, I don't have any HW to test this, nor use-cases for setting it up in a cloud environment.
I think mstpd is better suited for the embedded space; or that used to be the case 4-5 years ago

There are 2-3 people that may know the code better, but since this code is ~10 years old, I am not sure they remember stuff.
Most people have moved on.
And for the most part the current state of the project seems sufficient for most users.
There aren't many complaints from users.

Moving forward, I can add you ( @KanjiMonster ) to the group and you can drive this per-vlan stuff forward.
[That is: if you want]
I recognize you from the OpenWrt project; so I trust you to move things forward.

That's the best I can do right now :)

Feel free to reach me on my personal email if you want to discuss more stuff offline.

troglobit:
This is very interesting, and I've been meaning to test it out myself. I have access to devices with switch fabric (EspressoBin) that can run the latest Linux kernel (Buildroot), with bridge offloading.  I'll see if I get some time over the coming weekend, but that's just FYI, don't hold any merge for that.  If I find anything I'll post follow-up issues/patches.

linux_pvstate/driver_deps.c
KanjiMonster: Actually I completely merged this into the main mstpd (well bridge*) code, since just setting the per-vlan state isn't enough.

KanjiMonster:
After having thought about it, and looking at the standard more, I completely abandonded the driver_dep.c approach and started modifying mstpd main code to switch to the per vlan STP state, if available. I'm still in the process of making it work as expected, but in general the behavior I am working on now is:

1. on startup, try to subscribe to the RTNLGRP_BRVLAN, if succeded switch to new per vlan stp state (else use the bridge stp state)
2. When adding a port, get a list of vlans defined for interface, set them to blocking, and set the bridge stp state to forwarding
3. when setting a per-tree-port to a state, use the known vlans table to check which vlans we actually need to set (and set them)
4. being subscribed to RTNLGRP_BRVLAN, we get notified for creation/deletetion of vlans, and update the per interface vlan table accordingly
5. When we get a RTM_NEWVLAN notification, set the per-tree-port state for that vlan
6. When removing a port, do the opposite of 2.

Once I have it in a working state, I will update this PR with my (WIP) code. I'm currently working on with 2 resp. 6, the rest is implemented and working.

pavlix:
Hello @KanjiMonster and others!

Well, I am currently experimenting with mstpd, Linux 5.10 (LTS), OpenWRT, netifd and a hardware switch with MSTI state support. I've finally persuaded the kernel to accept a bridge/switch configuration defined in UCI as follows:

```
config device switch
        option name switch0
        option type bridge
        option mstp_region TestRegion

config bridge-msti
        option device switch0
        option mstid 1
        option fids "1 2 3"

config bridge-fid
        option device switch0
        option fid 1
        option vids "1 2 3"

config bridge-vlan
        option device switch0
        option vlan 1
        option ports "lan1 lan2"
```

I'm doing so using custom (currently unpublished) patches to the kernel, iproute2 and mstpd. My approach differs substantially from the description I found in the comments here. I think the main difference is that I assume mstpd knows its stuff well, respect its configuration, and try to pass it through the bridge RTNL layer down to the switch hardware.

In my experiments, I assume:

  * The mstpd daemon defines vid2fid, fid2mstid and mstid2state mappings and all we need to do is to pass all of this to the switch hardware.
  * Yet we still need to respect the whole kernel bridging/switchdev/DSA machinery in order to keep all the tools happy.
  * The switch hardware may not be able to handle per-VLAN state for all of the VLANs and therefore the indirection/grouping via MSTIs is mandatory.

My experimental implementation does:

  * My network initscript instructs the kernel to create the bridge device and configure all VLANs including their FID and MSTID numbers.
  * The script instructs mstpd with the same configuration as the kernel so that the daemon can do its job once the interfaces are activated.
  * It starts netifd that performs the actual activation of the network devices keeping the pre-existing kernel bridge configuration.
  * The main job of netifd currently is to configure all the other layers including bridged team devices and CPU-side VLAN and IP/DHCP configuration.
  * Depending on the state of the MSTI trees, the kernel MSTID state is updated.

I like the idea of tweaking mstpd main code rather than the driver deps as the bridge port state for RSTP is already handled there. It's great that we can now at least set the per-VLAN state in the kernel. However, this is not enough to get the configuration passed to the hardware that supports a limited number of per-MSTI state records.

In the pre-configuration phase I use the following commands to maintain the configuration:

```
mstpctl addbridge $name
mstpctl setforcevers $name mstp
mstpctl setmstconfid $name 0 $mstp_region
mstpctl createtree $device $mstid
mstpctl setfid2mstid $device $mstid:$fid
mstpctl setvid2fid $device $vid:$fid
bridge vlan add dev $port vid $vlan fid $fid sid $sid
```

The last command needs modification to kernel and iproute in order to route the vid/fid/mstid mapping all down to the device driver. Another modification is needed to route the mstid/state mapping for which my modified iproute2 allows the following command.

```
bridge vlan add dev $port sid $sid state $state
```

Again the request gets routed via RTNL to kernel and via switchdev to the target switch driver.

As long as you only want to create a single bridge over the ports of the switch, the vid/fid/mstid relation can be passed to the hardware unmodified. In case of multiple bridges over one switch, the driver would have to accept a virtual vid/fid/mstid mapping and translate that to a set of free fids and mstids. The switchdev interface allows for and easy way to report that a specific configuration is unsupported by the driver.

Would you be interested in cooperation to get the patches for the aforementioned approach ready and included in the respective projects? Please let me know.

commodo:
>  My approach differs substantially from the description I found in the comments here. I think the main difference is that I assume mstpd knows its stuff well, respect its configuration, and try to pass it through the bridge RTNL layer down to the switch hardware.

I can only say that it knows RSTP well. MSTP may need some testing for compatibility with other switches (HP, Cisco, Dell, etc).

> Would you be interested in cooperation to get the patches for the aforementioned approach ready and included in the respective projects? Please let me know.

Short answer: yes.

It seems like lately I am the one replying back to people.
Unfortunately I haven't been able to dig deep in the mstpd code since 2015 and I forgot how some details are done.

If you want, I can add you to the mstpd project group to be more directly involved.

KanjiMonster:
Sorry for taking so long (vacation got in between), I finally got around pushing our current code state (not cleaned up).

The build failure is because the builder doesn't seem to have recent enough kernel headers (i.e. 5.6+). This should probably be detectable (-> TODO).

The issue with directly talking with the switch asic is that using MSTP requires a switch asic that understands MSTP - using the per vlan stp states allows using MSTP also by bridging any network interfaces, even if they are not backed by any switching hardware.

That being said, if we do want to make use of MTSP support in switch asics, I think the kernel will need to know some basic knowledge about MST.

We are currently doing a per vlan MSTI internally when translating the per-vlan states to the switch chip, but switch chips tend to only support a limited amount of trees, especially since the standard limits it to 64 trees in total as the maximum.

So if we ever want to have proper DSA/switchdev support, we probably need to somehow store the semantic vlan -> MSTI mapping in a way that switch drivers have a chance to pick that up, without breaking the non-switch use case.

Having per-vlan STP states does have its advantage though, since this allows implementing PvSTP(+) compatibility.

KanjiMonster:
> I think the main difference is that I assume mstpd knows its stuff well, respect its configuration, and try to pass it through the bridge RTNL layer down to the switch hardware.

Our approach is being switch-agnostic, i.e, that mstpd does not have to know or care, or even should not care if the ports of the bridge are backed by a switch chip handling forwarding (and (M)STP) in hardware, or are just normal ethernet cards.

> In my experiments, I assume:
> 
>     * The mstpd daemon defines vid2fid, fid2mstid and mstid2state mappings and all we need to do is to pass all of this to the switch hardware.

As far as the linux networking stack is concerned, there currently is only IVL and no SVL.

To reflect this in mstpd I thought about making a static 1:1 mapping of FID to VID the default behaviour (so VID 1 is FID 1, VID 2 is FID 2 etc, with all FIDs assigned to the CIST by default of course).

This also has the (slight) advantage that you don't need to care about FIDs directly anymore, and can just directly assign VIDs (or rather their FIDs) to MSTIs. 

>     * The switch hardware may not be able to handle per-VLAN state for all of the VLANs and therefore the indirection/grouping via MSTIs is mandatory.

As I mentioned in my previous, this is indeed an issue and that semantic information needs to be stored in the kernel somewhere. Whether there is hardware implementing that mapping, or just a network card where linux will do the filtering in software should then not matter anymore for mstpd (or any other daemon trying to implement MSP). 
 
> 
> My experimental implementation does:
> 
>     * My network initscript instructs the kernel to create the bridge device and configure all VLANs including their FID and MSTID numbers.
> 
>     * The script instructs mstpd with the same configuration as the kernel so that the daemon can do its job once the interfaces are activated.
> 
>     * It starts netifd that performs the actual activation of the network devices keeping the pre-existing kernel bridge configuration.
> 
>     * The main job of netifd currently is to configure all the other layers including bridged team devices and CPU-side VLAN and IP/DHCP configuration.
> 
>     * Depending on the state of the MSTI trees, the kernel MSTID state is updated.
> 
> 
> I like the idea of tweaking mstpd main code rather than the driver deps as the bridge port state for RSTP is already handled there. It's great that we can now at least set the per-VLAN state in the kernel. However, this is not enough to get the configuration passed to the hardware that supports a limited number of per-MSTI state records.
> 
> In the pre-configuration phase I use the following commands to maintain the configuration:
> 
> ```
> mstpctl addbridge $device
> mstpctl setforcevers $device mstp
> mstpctl setmstconfid $device 0 $mstp_region
> mstpctl createtree $device $mstid
> mstpctl setfid2mstid $device $mstid:$fid
> mstpctl setvid2fid $device $vid:$fid
> bridge vlan add dev $port vid $vid fid $fid sid $sid
> ```
> 
> The last command needs modification to kernel and iproute in order to route the vid/fid/mstid mapping all down to the device driver. Another modification is needed to route the mstid/state mapping for which my modified iproute2 allows the following command.
> 
> ```
> bridge vlan add dev $port sid $sid state $state
> ```
> 
> Again the request gets routed via RTNL to kernel and via switchdev to the target switch driver.

On a first glance, this sounds like exactly what I am talking about above ;-). I assume any MSTI port states are then translated to individual Per-VLAN STP port states?

> As long as you only want to create a single bridge over the ports of the switch, the vid/fid/mstid relation can be passed to the hardware unmodified. In case of multiple bridges over one switch, the driver would have to accept a virtual vid/fid/mstid mapping and translate that to a set of free fids and mstids. The switchdev interface allows for and easy way to report that a specific configuration is unsupported by the driver.

I don't think switchdev or DSA support assigning ports to more than one bridge, so this can be probably ignored.

> Would you be interested in cooperation to get the patches for the aforementioned approach ready and included in the respective projects? Please let me know.

Definitely, especially if this includes upstreaming that stuff.

KanjiMonster:
> ```
> bridge vlan add dev $port vid $vid fid $fid sid $sid
> ```

What happens if you set different FID/SID (what does SID stand for? I would have expected e.g. TID as in Tree-ID). For different ports?

I think if we do want to get the mtsp information into the kernel, we should need to do something like

```
*enable MSTI grouping support in bridge* 
bridge vlan set-msti vid $vid fid $fid sid $sid
bridge vlan add dev $port vid $vid
```
 
> ```
> bridge vlan add dev $port sid $sid state $state
> ```

And similarly here 

```
bridge vlan set-state dev $port <vid $vid|sid $sid> state $state 
```

Setting the MSTI port state would then trigger a notification from the kernel that the port's MSTI state changed, and a bunch of VLANDB notifications for STP state changes (for all vlans of that MSTI that are present on that port).

pavlix:
> Our approach is being switch-agnostic.

The only difference between ways is that I am sharing more information with the kernel (as additional attributes to the VLAN) that can be either ignored (per-VLAN state) or used for the offload (per-MSTI state).

I'm looking forward to posting my patch for your review. It follow's @commodo 's request to have a more independent layer for the bridge/switch configuration.

> i.e, that mstpd does not have to know or care, or even should not care if the ports of the bridge are backed by a switch chip

It doesn't have to, as long as the required information is available to the kernel code.

> As far as the linux networking stack is concerned, there currently is only IVL and no SVL.

I'm working with a modified kernel. I will provide the MSTP-related modifications once ready.

> This also has the (slight) advantage that you don't need to care about FIDs directly anymore, and can just directly assign VIDs (or rather their FIDs) to MSTIs.

Agreed.

> As I mentioned in my previous, this is indeed an issue and that semantic information needs to be stored in the kernel somewhere.

There are a few possibilities. Let us see what can be done.

> Wouldn't it make more sense to let mstpd talk to the kernel directly in this case?

Of course! If only mstpd had all the information and netifd or another network manager can take over and do its own job. Right now mstpd doesn't seem to know which VLANs are active on which port. We both use `BRIDGE_VLAN_INFO_ONLY_OPTS` to avoid adding new VLANs to ports by mistake.

> So you don't need to modify iproute2.

Yep. I got rid of iproute2 modifications and now pass vid2mstid directly to the kernel.

> On a first glance, this sounds like exactly what I am talking about above ;-). I assume any MSTI port states are then translated to individual Per-VLAN STP port states?

Right now I'm just passing them to hardware. But yes, they should also translate to per-VLAN or so I guess, at least for all non-supporting drivers.

> I don't think switchdev or DSA support assigning ports to more than one bridge, so this can be probably ignored.

It is not forbidden in general. But the driver can complain. Let's ignore it for now.

> Definitely, especially if this includes upstreaming that stuff.

I guess this won't be easy to get through on the kernel side. Let's try.

> What happens if you set different FID/SID (what does SID stand for? I would have expected e.g. TID as in Tree-ID). For different ports?

I'm working with the mv88e6xxx driver. It uses VTU/STU and VID/SID as its terminology. I don't insist on that. The mappings in that driver are vid2fid and vid2sid rather than vid2fid2sid. I haven't given much of a look to other drivers.

> I think if we do want to get the mtsp information into the kernel, we should need to do something like
> 
> ```
> *enable MSTI grouping support in bridge* 
> bridge vlan set-msti vid $vid fid $fid sid $sid
> bridge vlan add dev $port vid $vid
> ```

That sounds good to me.

I'll post as a pull request. I'll probably start a common issue for the effort.

pavlix:
A few notes to the original questions.

> Is it correct that MSTP(d) expects a VLAN to be present/available on all bridge interfaces?

As far as I understand, MSTP (and mstpd) doesn't care. It maintains the CIST state of the ports and a set of MSTI states. You should be able to add/remove VLANs without disturbing mstpd.

> If so, then we need to make MSTPd listen for VLANDB entries and dynamically apply STP state if they get created, else STP state might get lost if VLANs get created after MSTPd applied the VLAN state of a MSTI.

Well, if kernel maintained the vid2mstid and mstid2state relations, then mstpd wouldn't even need to be involved.

> And more a understanding question of MSTP, since AFAICT all bridges in a MST are treated as one bridge for the CIST, what does that mean for the CIST port states within the region? Are they all forwarding, or should they be also be set to some non-looping state?

As far as I understand, you still use (C)IST to configure port states internal to the region while CST maintains the external relations that span over switches that don't even need to support MSTP but these still work together to give you the base layer for your MSTIs that are strictly regional.
 
> Looking at figure 13-11 of 802.11Q-2018 (p. 496), there on the CIST the port of bridge 84 to LAN J (to bridge 97) is blocked, but in the MSTI topology it is forwarding.

That is okay. The port is blocked for the CIST topology but unblocked for specific VLANs determined by the MSTI topology. Your per-VLAN state must get the chance to override the port state in order to make MSTP useful.

pavlix:
> If you want, I can add you to the mstpd project group to be more directly involved.

Thanks. Why not.

pavlix:
> The build failure is because the builder doesn't seem to have recent enough kernel headers (i.e. 5.6+). This should probably be detectable (-> TODO).

We'd better update the builder to the latest LTS kernel which is 5.10 if possible, so that we can actually check the builds rather than work around them.

KanjiMonster:
> > As far as the linux networking stack is concerned, there currently is only IVL and no SVL.
> 
> I'm working with a modified kernel. I will provide the MSTP-related modifications once ready.

So did you enhance linux' FDB code to support SVL? Because unless we do that, we probably should not try to program the switch to behave differently than linux expects it to/supports. Might lead to weird behavior if Linux learned a mac at different ports in different VLANs, but the switch has the mac on one of these ports for both VLANs because of SVL.

> > Wouldn't it make more sense to let mstpd talk to the kernel directly in this case?
> 
> Of course! If only mstpd had all the information and netifd or another network manager can take over and do its own job. Right now mstpd doesn't seem to know which VLANs are active on which port. We both use `BRIDGE_VLAN_INFO_ONLY_OPTS` to avoid adding new VLANs to ports by mistake.

See my code in this PR, I add that information to mstpd (as an array of STP states, where 0xff means not present) by listening to VLANDB notifications ;-)
 
> > So you don't need to modify iproute2.
> 
> Yep. I got rid of iproute2 modifications and now pass vid2mstid directly to the kernel.

Though of course having a utility for reading out the kernel state is always good (we have a modified iproute2 as well to check that mstpd is doing the expected thing).

> 
> > On a first glance, this sounds like exactly what I am talking about above ;-). I assume any MSTI port states are then translated to individual Per-VLAN STP port states?
> 
> Right now I'm just passing them to hardware. But yes, they should also translate to per-VLAN or so I guess, at least for all non-supporting drivers.
> 
> > I don't think switchdev or DSA support assigning ports to more than one bridge, so this can be probably ignored.
> 
> It is not forbidden in general. But the driver can complain. Let's ignore it for now.

Hmm, I thought I read that somewhere, but can't find it currently.

> > What happens if you set different FID/SID (what does SID stand for? I would have expected e.g. TID as in Tree-ID). For different ports?
> 
> I'm working with the mv88e6xxx driver. It uses VTU/STU and VID/SID as its terminology. I don't insist on that. The mappings in that driver are vid2fid and vid2sid rather than vid2fid2sid. I haven't given much of a look to other drivers.

We are working with Broadcom switches, these don't have any concept of FIDs at all. So for MSTI, you directly assign VIDs to trees (called STGs there). Luckily for us the more advanced ones support ~511 groups, so we don't run out of groups that fast.

> > I think if we do want to get the mtsp information into the kernel, we should need to do something like
> > ```
> > *enable MSTI grouping support in bridge* 
> > bridge vlan set-msti vid $vid fid $fid sid $sid
> > bridge vlan add dev $port vid $vid
> > ```
> 
> That sounds good to me.
> 
> I'll post as a pull request. I'll probably start a common issue for the effort.

Maybe similar to how we have a stp field for bridges whether to enable STP, we need a stp_mode for it as well to remain backward compatible.

Currently we have
* STP state per port (defaults to forwarding if stp=0, blocking if stp>0)
* per VLAN STP state defaults to forwarding on creation

This currently leads to the issue that you if you try to do per-vlan STP, the ports are (shortly) in forwarding mode when the vlan is added.

So I think we might want to have a new bridge sysfs option `stp_mode` variable which works like this:

stp_mode = 0 (default behaviour, "(R)STP") - as above

stp_mode = 1 (per vlan-stp)
* STP port state defaults to forwarding (so per vlan-state is affective)
* per VLAN STP state defaults to blocking on creation

stp_mode = 2 (MSTP)
* STP port state defaults to forwarding (so per vlan-state is affective)
* Default 
* You can create MSTIDs and assign VIDs to them (=> switchdev/dsa may reject creation if the table is full).
* per VLAN port STP state defaults to the MSTIs state for that port on creation
* setting a port's state for a MSTI will change the state for all VLAN STP states

This should allow both not breaking userspace by changing the current behavior, as well as allowing e.g. mstpd to detect at runtime what is supported (if the stp_mode sysfs option is present for the bridge, and if it can set the stp_mode to MSTP) and then work accordingly.

KanjiMonster:
> > The build failure is because the builder doesn't seem to have recent enough kernel headers (i.e. 5.6+). This should probably be detectable (-> TODO).
> 
> We'd better update the builder to the latest LTS kernel which is 5.10 if possible, so that we can actually check the builds rather than work around them.

Yes, though we probably don't want to require having the latest kernel headers available.

Maybe providing a local copy like several other projects do (like e.g. iproute2) would make sense.

pavlix:
> So did you enhance linux' FDB code to support SVL?

What exactly is SVL? Multiple FDB databases? For now I only made minimum changes to support the switch. Let's not call it a full implementation of the concept. 

> Because unless we do that, we probably should not try to program the switch to behave differently than linux expects it to/supports.

We can certainly delay that. But for my particular use case the switch is the priority and Linux networking is of less importance. So I first focus on getting the configuration in the switch rather than polishing the linux bridging layer.

> See my code in this PR, I add that information to mstpd (as an array of STP states, where 0xff means not present) by listening to VLANDB notifications ;-)

Well, you may compare our solutions now. Including the kernel patches to support the additional information.

> Though of course having a utility for reading out the kernel state is always good (we have a modified iproute2 as well to check that mstpd is doing the expected thing).

Right now I'm using mstpd logs, kernel logs and hardware tables to check that. But iproute2 support would be handy. Is that submitted upstream yet?

> We are working with Broadcom switches, these don't have any concept of FIDs at all. So for MSTI, you directly assign VIDs to trees (called STGs there).

So we both have direct relation between VID and MSTI but mv88e6xx supports FIDs that allow you to set up the right MAC tables for your VLANs.

> This currently leads to the issue that you if you try to do per-vlan STP, the ports are (shortly) in forwarding mode when the vlan is added.

Can't we just create the VLAN *with state assigned* right in the creation message, and then make sure the kernel avoids the race condition? After all it's the whole topology that's important, and MSTP is there mainly to optimize the flow so that you don't cancel one link for all VLANs while overusing the other links.

> This should allow both not breaking userspace by changing the current behavior.

The above way would fix the behavior for per-vlan-state-aware userland and keep behavior for per-vlan-state-agnostic userland.

> as well as allowing e.g. mstpd to detect at runtime what is supported

Well it would definitely be cool to detect support at runtime one way or another.

> Maybe providing a local copy like several other projects do (like e.g. iproute2) would make sense.

Sounds well once all kernel features are detected at runtime. Right now I'm using the headers to see whether I'm working with upstream kernel or experimental kernel with per-MSTI state support. This plays well with the OpenWRT build system. Runtime check would be much better, though.

KanjiMonster:
> > So did you enhance linux' FDB code to support SVL?
> 
> What exactly is SVL? Multiple FDB databases? For now I only made minimum changes to support the switch. Let's not call it a full implementation of the concept.

SVL = Shared Vlan Learning. So e.g. if you say VID 10 and VID 20 have the same FID, they share their learned mac addresses. 

Lets say you have two VIDs, 10 and 20. And you see a frame with mac 1:2:3:4:5 first at port 5 tagged with VID 10, then at port 6 tagged with VID 20.

With IVL (independend Vlan Learning), or each VID having a different FID, they will learn:
* VID 10 => FID A, VID 20 => FID B

* FID A: mac 1:2:3:4:5 is at port 5
* FID B: mac 1:2:3:4:5 is at port 6

With SVL, and both VIDs using the same FID, they will only have on entry:

* VID 10 => FID A, VID 20 => FID A
* FID A: mac 1:2:3:4:5 is at port 6

But Linux itself will always do IVL, regardless of what the switch ARL table does, so you will then have the situation:

Switch thinks mac 1:2:3:4:5 is at port 6 for VID 10 / FID A

Linux thinks mac 1:2:3:4:5 is at port 5 for VID 10 (because it does not support shared learning).

And since DSA will force the packets from linux itself destined to 1:2:3:4:5 VID 10 to port 5 (since this is what its FDB says), they go out the wrong port.

> > Because unless we do that, we probably should not try to program the switch to behave differently than linux expects it to/supports.
> 
> We can certainly delay that. But for my particular use case the switch is the priority and Linux networking is of less importance. So I first focus on getting the configuration in the switch rather than polishing the linux bridging layer.

My main point is that we should skip the FID part for now (or make IVL the default), until we can teach linux to also support it.

Assuming you have enough FIDs available for all VIDs.

> Right now I'm using mstpd logs, kernel logs and hardware tables to check that. But iproute2 support would be handy. Is that submitted upstream yet?

Not yet submitted upstream, but it can be found here https://github.com/rubensfig/iproute2-next/commit/839926c3ffcb09a69658fdf70bf0d6a7fbc22162

Our plan is to submit this upstream once we added setting the state as well.

> > We are working with Broadcom switches, these don't have any concept of FIDs at all. So for MSTI, you directly assign VIDs to trees (called STGs there).
> 
> So we both have direct relation between VID and MSTI but mv88e6xx supports FIDs that allow you to set up the right MAC tables for your VLANs.
> 
> > This currently leads to the issue that you if you try to do per-vlan STP, the ports are (shortly) in forwarding mode when the vlan is added.
> 
> Can't we just create the VLAN _with state assigned_ right in the creation message, and then make sure the kernel avoids the race condition? After all it's the whole topology that's important, and MSTP is there mainly to optimize the flow so that you don't cancel one link for all VLANs while overusing the other links.

Vlans are usually created by network managers (netifd, systemd-networkd, iproute2-bridge) that have no way of knowing the expected STP state for the vlans.

Unless you want to go around and patch all of these to get the state from somewhere, it might make more sense to make the kernel create the VLANs with a blocking state if vlan-aware STP is enabled, and let mstpd set the stp state for the vlans to something else.

Of course once you have grouped the vlans in MSTIs that should also work as just setting the MSTI's port state as the initial VLAN state, but currently there is no such thing (in upstream linux).

pavlix:
> SVL = Shared Vlan Learning. So e.g. if you say VID 10 and VID 20 have the same FID, they share their learned mac addresses.

This is what the hardware does. Sharing the tables sounds easy to achieve. It is splitting that might be complicated. Sometimes I wonder if it would be better to leave this completely up to the switch to **learn** the addresses and then just pass them to the bridge. But I haven't dug so deep, so maybe this idea is completely bogus.

> Lets say you have two VIDs, 10 and 20. And you see a frame with mac 1:2:3:4:5 first at port 5 tagged with VID 10, then at port 6 tagged with VID 20.

Ah, yes. Then both hardware and software should *only* share the record betweeen VIDs configured with the same FID. However, independent should always be good enough, just suboptimal, right?

> But Linux itself will always do IVL, regardless of what the switch ARL table does, so you will then have the situation:
> 
> Switch thinks mac 1:2:3:4:5 is at port 6 for VID 10 / FID A
> 
> Linux thinks mac 1:2:3:4:5 is at port 5 for VID 10 (because it does not support shared learning).
> 
> And since DSA will force the packets from linux itself destined to 1:2:3:4:5 VID 10 to port 5 (since this is what its FDB says), they go out the wrong port.

Well I would expect that IVL is per-VID and therefore learns independently for VLAN 10 and VLAN 20, while SVL is per-VID, and therefore could share the records *if* VLAN 10 and VLAN 20 belong to the same FID. Care to explain?

Plus I would really like DSA to pass all the traffic right to the switch and let it do its job rather than distribute the frames and send them over specific ports. Especially when it comes to multicast, I'd love to avoid sending duplicate frames through the CPU port.

> My main point is that we should skip the FID part for now (or make IVL the default), until we can teach linux to also support it.

I'm basically ignoring it already because it's not a priority for me at all, just wanted to have it in the design.

> Assuming you have enough FIDs available for all VIDs.

Ack.

> > Right now I'm using mstpd logs, kernel logs and hardware tables to check that. But iproute2 support would be handy. Is that submitted upstream yet?
> 
> Not yet submitted upstream, but it can be found here [rubensfig/iproute2-next@839926c](https://github.com/rubensfig/iproute2-next/commit/839926c3ffcb09a69658fdf70bf0d6a7fbc22162)

Thanks! I'll add it to my patchset to experiment for now.

> Vlans are usually created by network managers (netifd, systemd-networkd, iproute2-bridge) that have no way of knowing the expected STP state for the vlans.

I do not see a difference between port STP and vlan STP in this regard. Both ports and vlans are added/activated by the network manager and both should IMO start in a safe mode that can be taken over by mstpd.

I am using the same source of configuration for both *netifd* and *mstpd* exactly so that the configuration *always* matches. My experience suggests that mstpd configuration would better be integrated into the main configuration mechanism.

> Unless you want to go around and patch all of these to get the state from somewhere, it might make more sense to make the kernel create the VLANs with a blocking state if vlan-aware STP is enabled, and let mstpd set the stp state for the vlans to something else.

I might be wrong but I feel okay patching the network manager bridging always integrate correctly with mstpd and maintain a consistent configuration. After all I treat mstpd as sort of a userland bridge driver, together with the bridging code in the network manager.

> Of course once you have grouped the vlans in MSTIs that should also work as just setting the MSTI's port state as the initial VLAN state, but currently there is no such thing (in upstream linux).

Well, yes, this is yet another reason I like the idea of giving the kernel the proper mstid2state configuration and then use it for new VLANs. That way mstpd could pre-configure everything in time.

KanjiMonster:
> > SVL = Shared Vlan Learning. So e.g. if you say VID 10 and VID 20 have the same FID, they share their learned mac addresses.
> 
> This is what the hardware does. Sharing the tables sounds easy to achieve. It is splitting that might be complicated. Sometimes I wonder if it would be better to leave this completely up to the switch to **learn** the addresses and then just pass them to the bridge. But I haven't dug so deep, so maybe this idea is completely bogus.

Sharing the tables isn't easy if sharing isn't supported (e.g. Broadcom). But anyway, the linux networking stack must match what the switch does, else you will get hard to debug issues. And in the same way the switch should not be configured for things the Linux networking stack does not support. So if the Linux does not support Shared Vlan Learning, then we should avoid doing it in the switch chip.

> > Lets say you have two VIDs, 10 and 20. And you see a frame with mac 1:2:3:4:5 first at port 5 tagged with VID 10, then at port 6 tagged with VID 20.
> 
> Ah, yes. Then both hardware and software should _only_ share the record betweeen VIDs configured with the same FID. However, independent should always be good enough, just suboptimal, right?

The important part is that both the switch and linux agree on how the ARL table / learned addresses look like.

> > But Linux itself will always do IVL, regardless of what the switch ARL table does, so you will then have the situation:
> > Switch thinks mac 1:2:3:4:5 is at port 6 for VID 10 / FID A
> > Linux thinks mac 1:2:3:4:5 is at port 5 for VID 10 (because it does not support shared learning).
> > And since DSA will force the packets from linux itself destined to 1:2:3:4:5 VID 10 to port 5 (since this is what its FDB says), they go out the wrong port.
> 
> Well I would expect that IVL is per-VID and therefore learns independently for VLAN 10 and VLAN 20, while SVL is per-VID, and therefore could share the records _if_ VLAN 10 and VLAN 20 belong to the same FID. Care to explain?

Yes. I was just describing what happens if the switch chip does SVL, while Linux does IVL.

> Plus I would really like DSA to pass all the traffic right to the switch and let it do its job rather than distribute the frames and send them over specific ports. Especially when it comes to multicast, I'd love to avoid sending duplicate frames through the CPU port.

Sure, but sometimes services on the switch will need to communicate with the outside (e.g. for IGMP/MLD snooping, DHCP, etc), so Linux should also have a correct view of what is connected to the switch's ports.

> 
> > My main point is that we should skip the FID part for now (or make IVL the default), until we can teach linux to also support it.
> 
> I'm basically ignoring it already because it's not a priority for me at all, just wanted to have it in the design.
> 
> > Assuming you have enough FIDs available for all VIDs.
> 
> Ack.
> 
> > > Right now I'm using mstpd logs, kernel logs and hardware tables to check that. But iproute2 support would be handy. Is that submitted upstream yet?
> > 
> > 
> > Not yet submitted upstream, but it can be found here [rubensfig/iproute2-next@839926c](https://github.com/rubensfig/iproute2-next/commit/839926c3ffcb09a69658fdf70bf0d6a7fbc22162)
> 
> Thanks! I'll add it to my patchset to experiment for now.
> 
> > Vlans are usually created by network managers (netifd, systemd-networkd, iproute2-bridge) that have no way of knowing the expected STP state for the vlans.
> 
> I do not see a difference between port STP and vlan STP in this regard. Both ports and vlans are added/activated by the network manager and both should IMO start in a safe mode that can be taken over by mstpd.

The way the per-VLAN STP state support added in the linux kernel is (slightly simplified):

* by default all vlans created are in fordwarding state
* If the per-port STP state is blocking, all VLANs are treated as blocking
* If the per-port STP state is forwarding, the VLAN state is observed

I assume this is done to not break userspace, because if you started the vlans in blocked state, this would break all current STP userspace implementations, as they don't know how to unblock vlans.

In the same vein, vlans in forwarding cannot override a port's blocked state, else current STP userspace implementations would be unable to block anything, since they don't know how to block vlans.

This is why I suggest an option for the bridge to tell it that the userspace application is vlan aware, and new vlans should be created in a safe state (i.e. blocking). Be it named `stp_mode`, `stp_vlan_mode`, `stp_vlan_aware` or however.

This mirrors how the kernel behaves when enabling stp on a bridge; it will set all ports to blocking/disabled by default (depending if they have a link), so if the stp implementation is vlan aware, it should work similar except per-vlan.

> I am using the same source of configuration for both _netifd_ and _mstpd_ exactly so that the configuration _always_ matches. My experience suggests that mstpd configuration would better be integrated into the main configuration mechanism.
> 
> > Unless you want to go around and patch all of these to get the state from somewhere, it might make more sense to make the kernel create the VLANs with a blocking state if vlan-aware STP is enabled, and let mstpd set the stp state for the vlans to something else.
> 
> I might be wrong but I feel okay patching the network manager bridging always integrate correctly with mstpd and maintain a consistent configuration. After all I treat mstpd as sort of a userland bridge driver, together with the bridging code in the network manager.

I don't think any network manager sets any (initial) STP states for ports, so I fail to see why they now should need to do that for MSTP. Setting VLAN -> (FID ->) MSTID tables sure, but actual states are the role of the STP implementation that also processes the BPDUs and know the bridge topology.

Also there's always the option that there is no network manager, and the bridge (and mstpd) is setup by a bunch of shell scripts.

> > Of course once you have grouped the vlans in MSTIs that should also work as just setting the MSTI's port state as the initial VLAN state, but currently there is no such thing (in upstream linux).
> 
> Well, yes, this is yet another reason I like the idea of giving the kernel the proper mstid2state configuration and then use it for new VLANs. That way mstpd could pre-configure everything in time.

pavlix:
> Sharing the tables isn't easy if sharing isn't supported (e.g. Broadcom).

Unless I already said that, your help will be crucial in this.

> And in the same way the switch should not be configured for things the Linux networking stack does not support.

I do not have that luxury. I don't mind making it optional and clearly mark as experimental but I *need* to configure the switch as it is supposed to be configured and then hopefully catch up with the Linux kernel part. Let's make the software as useful as possible even if it means some features have to be carefully enabled at build time.

> So if the Linux does not support Shared Vlan Learning, then we should avoid doing it in the switch chip.

I can't afford to randomly dump features just because Linux doesn't support them yet. The only other way for me is to keep custom private patches. I've been in open source for a long time and I don't believe private patches are the best approach.

> Yes. I was just describing what happens if the switch chip does SVL, while Linux does IVL.

Well I do not see the problem yet. Maybe have an interactive chat or call some day?

> Sure, but sometimes services on the switch will need to communicate with the outside (e.g. for IGMP/MLD snooping, DHCP, etc), so Linux should also have a correct view of what is connected to the switch's ports.

Yes, we need to distinguish **forward** and **management** traffic. Those are two very different cases.

> The way the per-VLAN STP state support added in the linux kernel is (slightly simplified):
> 
>     * by default all vlans created are in fordwarding state

One way or another, we'll need to start them safe. Maybe a global or per-bridge knob will indeed be necessary to trigger the right behavior.

>     * If the per-port STP state is blocking, all VLANs are treated as blocking

If that is the case, this needs to be changed.

> I assume this is done to not break userspace, because if you started the vlans in blocked state, this would break all current STP userspace implementations, as they don't know how to unblock vlans.

Well it's a good default in case you don't set per-VLAN states at all. If you set set a port+vlan to *forwarding* then you probably mean it and it should take precedence over the blocking port state. After all it is the whole point of MSTP to “route” VLANs differently.

> In the same vein, vlans in forwarding cannot override a port's blocked state, else current STP userspace implementations would be unable to block anything, since they don't know how to block vlans.

A correct implementation would IMO start VLANs in *follow-the-port-state* mode by default and *disabled/blocking* in MSTP, PVST+ and similar cases.

> This is why I suggest an option for the bridge to tell it that the userspace application is vlan aware, and new vlans should be created in a safe state (i.e. blocking). Be it named `stp_mode`, `stp_vlan_mode`, `stp_vlan_aware` or however.

Looks like there will be no other way to fix the kernel otherwise. So we'll probably have to propose the *bridge-state-port-only*, *bridge-state-per-vlan* and *bridge-state-per-msti* ternary option or something along that line just as you suggested. That would also solve your concern about the differences between per-VLAN and per-MSTI states. You'd have to configure (or detect) what's to be done.

> I don't think any network manager sets any (initial) STP states for ports, so I fail to see why they now should need to do that for MSTP.

Fair enough. The integration might even be easier of the kernel knob is triggered at mstpd start long before the manager has a chance to make any configuration happen. On the other hand, I'm thinking in a compact system were I have all the tools under control.

> Setting VLAN -> (FID ->) MSTID tables sure, but actual states are the role of the STP implementation that also processes the BPDUs and know the bridge topology.

Well in this case you could actually avoid this if you wanted to and just set the mapping in mstpd and VLAN assignment in the network manager. This is how it works now and sounds like a good start.

> Also there's always the option that there is no network manager, and the bridge (and mstpd) is setup by a bunch of shell scripts.

I don't really distinguish between a bunch of scripts and a network manager. I worked with ifcfg, NetworkManager, connman, wicked, OpenWRT network scripts, netifd… I treat them all the same. A bunch of scripts combined with a link monitor, a DHCP client and more does exactly the same job. Let's call it a network manager whether it's an actual management daemon or not.

Thanks a lot for your insights.

One thing… Currently I'm using the following three-phase solution for my experiments:

1) Create the bridge interface. Add it to mstpd. Configure the mstpd bridge from the UCI configuration. At this phase it is appropriate to set *per-msti-state* or *per-vlan-state* option to override the default behavior.

2) Start netifd. Use the same configuration source. It takes over the created bridge and does all the other configuration options. It applies other network configuration like adding virtual bridge vlan interfaces and IP configuration on top of that.

3) Now you can check what happens when you plug/unplug cables, make loops and more.

It looks clean enough to me for the beginning. What do you think?

pavlix:
@shockingchen Were you successful?

KanjiMonster:
> when I download the [53f1e61](https://github.com/mstpd/mstpd/commit/53f1e616d3e4f34d0f321d0371aa1bdce0a82689), build it on the ubuntu 18,it will throws errors,is there any update about this?

You need to install a newer kernel, 5.7 or newer. I'm not sure if there is one even available for ubuntu 18.04, you might need to upgrade to 20.4.

You also need to be running the newer kernel for the feature to be available.

KanjiMonster:
> > when I download the [53f1e61](https://github.com/mstpd/mstpd/commit/53f1e616d3e4f34d0f321d0371aa1bdce0a82689), build it on the ubuntu 18,it will throws errors,is there any update about this?
> 
> You need to install a newer kernel, 5.7 or newer. I'm not sure if there is one even available for ubuntu 18.04, you might need to upgrade to 20.4.
> 
> You also need to be running the newer kernel for the feature to be available.

After a bit more investigation, there actually is no reasonable way to do this on ubuntu 20.04 (or earlier), since the uapi headers are contained in a different package, libc-kernel-dev (or something like that), which is kept at the base kernel version level (e.g. 5.4), and there is no way to update the package except by going to a newer ubuntu version.

So I decided to add the required uapi headers to the mstpd package to allow building it on ubuntu 20.04. For now only strictly what is needed (rtnetlink.h for RTM_*VLAN and if_bridge.h for the appropriate structs for that).

Making use of the per vlan stp of course still requires a newer kernel, but that is easily doable through the hwe packages.

KanjiMonster:
I created an alternative implementation using the kernel mst support, which should also play nice with switchdev/DSA at https://github.com/mstpd/mstpd/pull/150.

PR #107, opened 2021-02-19
build: move -Werror flag under --enable-devel opt

commodo:
This change disables the -Werror flag under normal build. When updating
compilers, new warnings get enabled by default.
We will do this only during CI where the --enable-devel flag will be set.


commodo:
Will merge this in 1 day if no objections.

SoapGentoo (approved):
@commodo thanks for this, we have a followup commit where we'd like to disable pre-compressed manpages too, could we get that in too?

commodo:
sure;

PR #108, opened 2021-02-19
Don't compress man pages

SoapGentoo:
* Every distro and package manager uses different compression for
  man pages.


commodo (approved):

commodo:
will leave it open for 1 day in case there are comments;

PR #112, opened 2021-03-17
[RFC] independent linux bridging support module

pavlix:
This is how a separate bridge platform module could look like. Linux code is based on #103 and extended to support hardware offload with experimental changes in the Linux kernel.


KanjiMonster:
Where do I find your kernel changes?

commodo:
maybe let's start with the first 2 commits;
i.e.:
* libnetlink: introduce rta_nest and u8, u16, u64 helpers for nesting within rtattr
* libnetlink: Set NLA_F_NESTED in rta_nest 

maybe in a separate PR;
those 2 seem fine on their own

pavlix:
> Where do I find your kernel changes?

Commited to a contrib directory:

https://github.com/mstpd/mstpd/pull/112/commits/24ada5f0fa9a62177849df984279875f07ca96dd

Might be more convenient to pull the branch to see them. I hope they don't depend on any of my other changes and cleanly apply to any recent 5.10.y.

pavlix:
@KanjiMonster Could you please check that the non-experimental path includes all that we need? I think we can live with a half-baked experimental part since it is only triggered by a long experimental macro variable from the experimental kernel patches. These can be improved over the time and submitted to net-next when ready.

platform/linux-bridge.c
```diff
@@ -0,0 +1,141 @@
...
+#endif
+}
+
+int bridge_port_tree_set_state(unsigned ifindex, __u16 mstid, __u8 state)
+{
+    INFO("[linux-bridge] port %d mstid %d state %d", ifindex, mstid, state);
+    if (mstid == 0)
+        return port_update(ifindex, state);
+#ifdef RTM_GETVLAN
+#ifdef HAVE_EXPERIMENTAL_KERNEL_VID_FID_SID_STATE_RELATION
```
commodo, line 127: curios: where will these get defined?

if possible i would not add compiletime behavior for mstpd;
i am a bit out of touch with things, but could we do a runtime detection of these features?
maybe have a global variable  `static bool g_rtm_get_vlan_supported`?

there are are a few cases [don't know how many] where
1. some people prefer to backport patches onto kernel 5.4 [let's say] because it's an LTS kernel
2. some people may want to run mstpd on older kernels where these RTNL parameters get passed; although the kernel ignores then
  * if the kernel would ignore these anyway, could we maybe incorporate it in the mstpd behavior somehow to detect at runtime?

contrib/linux/patches-5.10/0001-switchdev-add-fid-sid-state-to-VLAN-notification.patch
```diff
@@ -0,0 +1,117 @@
+From 27dd8764c2c9ada4d9de83ae8d9d242418286b11 Mon Sep 17 00:00:00 2001
```
commodo, line 1: curios: these patches are meant as reference for others?
we usually didn't keep kernel patches in mstpd;

commodo:
> P. S. We'll need to upgrade the builder, ideally to a distro the latest LTS kernel which is 5.10.

we can always run whatever docker we want;
but if we need a newer builder for it's kernel, i would try to maybe define things in userspace [same as in kernelspace] to keep things a bit detached;

platform/linux-bridge.c
KanjiMonster: > curios: where will these get defined?
> 
> if possible i would not add compiletime behavior for mstpd;
> i am a bit out of touch with things, but could we do a runtime detection of these features?
> maybe have a global variable `static bool g_rtm_get_vlan_supported`?
> 
> there are are a few cases [don't know how many] where
> 
>     1. some people prefer to backport patches onto kernel 5.4 [let's say] because it's an LTS kernel
> 
>     2. some people may want to run mstpd on older kernels where these RTNL parameters get passed; although the kernel ignores then
> 
> 
>     * if the kernel would ignore these anyway, could we maybe incorporate it in the mstpd behavior somehow to detect at runtime?

This is what our code is doing here https://github.com/mstpd/mstpd/pull/103/files#diff-ec925e194e98ee3a76396b491ac94d2715496852feb14bcfdddeff4f9df8f381R322

Try to subscribe to the RTNLGRP_BRVLAN netlink notifications, and if it fails, assume per-vlan stp support is not available and fall back to the old behavior. So you only need newer headers to compile, but not run (which is why just having our own headers might be okay).

You obviously need the newer kernel to use it though (or in general any kernel that has the appropate per-vlan-state support).

platform/linux-bridge.c
```diff
@@ -0,0 +1,141 @@
...
+#else
+    return -ENOTSUP;
+#endif
+}
+
+int bridge_port_tree_set_state(unsigned ifindex, __u16 mstid, __u8 state)
+{
+    INFO("[linux-bridge] port %d mstid %d state %d", ifindex, mstid, state);
+    if (mstid == 0)
+        return port_update(ifindex, state);
```
KanjiMonster, line 125: This will break the linux side when the CIST is set to blocking, as now you will block all vlans.

See https://github.com/mstpd/mstpd/pull/103/commits/55985dc99d900bca1e89f942d5188ee002bbdcd4 , https://github.com/mstpd/mstpd/pull/103/commits/97cc2925cde383f8f0bf79ed6b9a228f0eead584 and https://github.com/mstpd/mstpd/pull/103/commits/53f1e616d3e4f34d0f321d0371aa1bdce0a82689 .

platform/linux-bridge.c
```diff
@@ -0,0 +1,141 @@
...
+#ifdef RTM_GETVLAN
+#ifdef HAVE_EXPERIMENTAL_KERNEL_VID_FID_SID_STATE_RELATION
+    return vlan_update(ifindex, 0, 0, mstid, state);
+#else
+    int vid;
+    int ret = 0;
+
+    for (vid = 1; ret == 0 && vid < VID_COUNT; vid++)
+        if (vid2mstid[vid] == mstid)
+            ret = vlan_update(ifindex, vid, state);
```
KanjiMonster, line 135: the kernel will return an error when you try to set the state of a vlan that isn't configured on a port, so you will most likely get a lot of errors here.

platform/linux-bridge.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+int bridge_port_tree_set_state(unsigned ifindex, __u16 mstid, __u8 state)
+{
+    INFO("[linux-bridge] port %d mstid %d state %d", ifindex, mstid, state);
+    if (mstid == 0)
+        return port_update(ifindex, state);
+#ifdef RTM_GETVLAN
+#ifdef HAVE_EXPERIMENTAL_KERNEL_VID_FID_SID_STATE_RELATION
+    return vlan_update(ifindex, 0, 0, mstid, state);
```
KanjiMonster, line 128: Where do you configure the MSTID => (FID?) => VID relation on the kernel? How will the kernel know which VIDs to block or unblock?

platform/linux-bridge.c
```diff
@@ -0,0 +1,141 @@
...
+
+    addraw_l(&req.n, sizeof(req.buf), RTA_DATA(rta), RTA_PAYLOAD(rta));
+
+    return rtnl_talk(rth, &req.n, 0, 0, NULL, NULL, NULL);
+}
+#endif
+
+#ifdef HAVE_EXPERIMENTAL_KERNEL_VID_FID_SID_STATE_RELATION
+#else
+__u16 vid2mstid[VID_COUNT] = {};
```
KanjiMonster, line 103: Please don't use static variables for storing state, add it to the appropriate bridge/port/etc structs.

pavlix: This is intentional and it follows @commodo's suggestion to turn the platform code into a library (in future) fully independent of any mstpd code. This relation store is owned by the platform library rather than mstpd.

Is there any use case where this might be an issue?

platform/linux-bridge.c
pavlix: This, in my opinion, will have to be fixed in the kernel. What is the best solution until then? Always opening the port will create switch loops, right?

platform/linux-bridge.c
pavlix: This is the mstid-based variant guarded by `HAVE_EXPERIMENTAL_KERNEL_VID_FID_SID_STATE_RELATION`. It won't be applied to kernels without the experimental patch at all. The choice is done at compile time.

platform/linux-bridge.c
pavlix: Should we then just ignore the errors here? Otherwise we'd have to try to keep track of the per-port VLAN configuration that sounds a little bit tedious but might be a good later addition.

pavlix:
> curios: where will these get defined?

`RTM_GETLVAN` is defined in the vanilla kernel since the addition of per-vlan bridge information.

`HAVE_EXPERIMENTAL_KERNEL_VID_FID_SID_STATE_RELATION` is only defined in the experimental patches attached to this pull request.

> maybe have a global variable static bool g_rtm_get_vlan_supported?

This part doesn't seem to be open to comments. I'm all for having runtime detection. Could we possibly first merge the framework to get some base code to work on… I mean to get the basic feature set for testing. And then make modifications and add runtime detection where applicable.

platform/linux-bridge.c
commodo: i guess there is a confusion;
@KanjiMonster is correct here; you are keeping this information globally for all bridges/switches/ports/etc

so, if you add 2 bridges/switches to mstpd, how would you know which `vid2mstid` corresponds to which switch/bridge?

this would maybe be better somewhere on the `bridge_t` struct for starter;

platform/linux-bridge.c
commodo: can we get commit https://github.com/mstpd/mstpd/pull/103/commits/55985dc99d900bca1e89f942d5188ee002bbdcd4   into a separate PR for review?

i know this may be tedious;
but at least it may be one way forward;
i like the idea of the mechanism of trying to subscribe to RTNLGRP_BRVLAN netlink notifications [with failure handling];

maybe after that, @KanjiMonster can you rebase PR https://github.com/mstpd/mstpd/pull/103/ ? 
if the commits go down in that PR, then it makes things easier [for me] to track;

pavlix:
> hello when will this feature support on the release version?

Hello! Nice to hear from someone interested in the topic. We experimented a bit, each if one in a slightly different way. We do appreciate any help, testing and improvements.

Bartel-C8:
Could the linux kernel changes be superseded by this merge to mainline kernel? (Yet to be released in kernel 6.0)

https://github.com/torvalds/linux/commit/82e94d4144d7a29e6e955e4b2ea681ed3f16d689

(and possibly https://github.com/torvalds/linux/commit/ec7328b59176227216c461601c6bd0e922232a9b )

commodo:
> Could the linux kernel changes be superseded by this merge to mainline kernel? (Yet to be released in kernel 6.0)
> 
> [torvalds/linux@82e94d4](https://github.com/torvalds/linux/commit/82e94d4144d7a29e6e955e4b2ea681ed3f16d689)
> 
> (and possibly [torvalds/linux@ec7328b](https://github.com/torvalds/linux/commit/ec7328b59176227216c461601c6bd0e922232a9b) )

Looks like it.
That's good.
This project hasn't been too active, and the MSTP part is not very robust (there are some issue with it).

Maybe we can start closing the project slowly, unless someone wants to pick it up :)

GrumpyTeaching:
> This has been in the kernel since 5.18 actually.

Looks like we can experiment with the current kernels, right?

pavlix:
> This has been in the kernel since 5.18 actually.

Looks like we could experiment with a current kernel, right?

KanjiMonster:
> > This has been in the kernel since 5.18 actually.
> 
> Looks like we could experiment with a current kernel, right?

Yes. I actually already did so while adding support for 6.1 to our platform.

My initial observations are:

- it can only be enabled when there are no VLANs configured on the bridge, so ideally you enable it while creating the bridge (mstpd trying to enable it may be too late)
- you can only assign existing VLANs to an MST, so you can't proactively e.g. assign VIDs 100-300 to MSTI 1
- there are no netlink notifications to userspace about port MST STP state changes, neither the actual MST STP state change, nor the triggered VLAN STP state changes

The last point currently makes it unusable for us, since we rely on the notifications to write the STP state to the ASIC, and we couldn't even listen to the VLAN STP changes as a work around. Nothing unsolvable, just more work.

There are kernel internal notifications though, so switchdev drivers would be able to implement it.

There are also a few open questions for me:

- Now that you can assign VIDs to MSTIs in the kernel, does it still make sense to configure the VID (-> FDB) -> MSTI assignment in mstpd, or should rather mstpd use the kernel VID -> MSTI configuration? The latter would mean you could configure the VIDs and their MSTIs with your favourite network manager, and it wouldn't need to know anything about mstpd. But it would mean that you would need to configure every VID on the bridge, since they would default to CIST / MSTI 0 when unconfigured.
- And more an implementation question, since MSTI configuration is tied to VIDs existing on the bridge, removing a VID will move it to MSTI 0, even when recreating it later. But AFAICT there is no notification telling userspace about that, so mstpd will need to do refcounting to know when a VID gets removed, and set the MSTI again when the VID reappears (or move the VID to CIST on deletion).

Unless someone beats me to it (and everyone is welcome to), I will probably start working on this early next year.

newsages (approved):

PR #113, opened 2021-03-17
libnetlink changes required for any VLAN/MSTI state solution

pavlix:
As requested in #112, these two patches that come from #103 can already be applied to support any future solution and experiments in setting the state for either a VLAN or an MSTI in Linux.

### DIFF
--- a/libnetlink.c
+++ b/libnetlink.c
@@ -611,6 +611,37 @@ int rta_addattr_l(struct rtattr *rta, int maxlen, int type,
 	return 0;
 }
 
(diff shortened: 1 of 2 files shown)


commodo (approved):

commodo:
will merge it today if no objections

commodo:
fyi: there should be an invitation to join the group

commodo:
and i've also opened an #mstpd  channel on freenode if that helps;
i did not setup any irc bot to keep it alive all the time; but i can login during work my hours;

PR #115, opened 2021-03-20
build with Ubuntu 20.04

pavlix:
Any changes that rely on new kernel features require either to switch to new kernel headers in the build system or to add the kernel headers to the project. I vote for using the distribution kernel headers rather than bundling them.

It provides better support for experimental builds on systems with experimental kernels used e.g. to test experimental features in #112.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ on: [ push, pull_request ]
 
 jobs:
   linux:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
     - name: build


pavlix:
Well, the kernel in 20.04 isn't new enough for new stuff, we'll do with the default for now and switch later, possibly building for both older and newer kernels.

.github/workflows/build.yml
```diff
@@ -4,7 +4,7 @@ on: [ push, pull_request ]
 
 jobs:
   linux:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
commodo, line 7: so, `ubuntu-latest` should be `ubuntu 20.04` soon anyway [if it isn't already];
but it is usually the LTS that most CIs use;

PR #118, opened 2021-05-25
utils/ifquery: drop python2 support

jklare:
* python2 is not maintained anymore and all scripts using it should be
  updated to python3
* ran 2to3 for utils/ifquery and added python3 shebang
* removed python3 sed from rpm.spec

Signed-off-by: Jan Klare <jan.klare@bisdn.de>

### DIFF
--- a/rpm.spec
+++ b/rpm.spec
@@ -34,12 +34,6 @@ make %{?_smp_mflags}
 rm -rf %{buildroot}
 make install DESTDIR=%{buildroot}
 
-# RHEL7/CentOS7 still permits #!/usr/bin/python in scripts.  Other supported
-# RPM-based distros/versions require #!/usr/bin/python3 in all scripts.
-%if !0%{?rhel} || 0%{?rhel} >= 8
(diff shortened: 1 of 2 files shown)


commodo (approved):

commodo:
LGTM

Will merge in 1-2 days.
In case someone has some complaints.

Related to this, there was an idea to remove `ifquery` completely and try to work without it.
If you feel this would be also interesting to take a look at, feel free :)

Danke
Alex

PaulSD:
LGTM
(I wrote the original code for this)

For reference:
The purpose of the ifquery script is to allow the existing ifupdown.sh and mstp_config_bridge scripts (which were designed for Debian-based distros) to also be used on RedHat-based distros.  The existing scripts will not work on RedHat systems without this ifquery script (this script is necessary to parse the /etc/syconfig/network-scripts/bridge-stp file).  To eliminate this script, someone would either need to write additional RedHat-specific versions of ifupdown.sh and mstp_config_bridge, or we would need to significantly rewrite the existing scripts to make them more directly compatible with both platforms.

PR #124, opened 2021-06-09
make: Updated Makefile.am to support distcheck

sshambar:
Updated Makefile.am to allow the make distcheck to work, which
means make install can be reversed by make uninstall, and
make clean correctly removes generated files.  make dist now
also creates a tarfile that can be used to configure and
build from.  Added README to install.

Signed-off-by: Scott Shambarger <devel@shambarger.net>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,10 +3,14 @@ ACLOCAL_AMFLAGS = -I m4
 sbin_PROGRAMS = mstpd mstpctl
 
 mstpd_SOURCES = \
-	main.c epoll_loop.c brmon.c bridge_track.c libnetlink.c mstp.c \
-	packet.c netif_utils.c ctl_socket_server.c hmac_md5.c driver_deps.c
+	main.c epoll_loop.c epoll_loop.h clock_gettime.h brmon.c \
+	bridge_track.c bridge_track.h driver.h bridge_ctl.h libnetlink.c \
+	libnetlink.h mstp.c mstp.h packet.c packet.h netif_utils.c \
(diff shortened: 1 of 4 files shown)


commodo:
is this adding new build rules to `make`
if yes, can we add some CI calls to this in the [build.yml file](https://github.com/mstpd/mstpd/blob/master/.github/workflows/build.yml)

i can add them, if you tell me the make commands

sshambar:
Well, it's not adding a target so much as fixing one :)  I added a CI entry for "make distcheck" and it passed on my fork.

I also added LICENSE to dist, and removed a couple generated files that shouldn't have been in the tar.

Makefile.am
```diff
@@ -102,7 +128,7 @@ rpm: rpm.spec
 	git clone "$$SrcDir" "mstpd-$$Version" ; \
 	tar -cvjf "mstpd-$$Version.tar.bz2" "mstpd-$$Version" ; \
 	rm -rf "mstpd-$$Version" ; \
-	cd .. ; mkdir -p SPECS ; cp "$$SrcDir/rpm.spec" SPECS/mstpd.spec
+	cd .. ; mkdir -p SPECS ; cp "$$SrcDir/rpm.spec" SPECS/mstpd.spec; \
```
PaulSD, line 131: This will cause a merge conflict with #119 ... You might want to skip it here to avoid that.

PaulSD (approved):
LGTM

Makefile.am
sshambar: Cool... was just fixing bug in the file (and perhaps 119 wouldn't be merged :), reverted the fix.

Issue #126, opened 2021-09-23
Check for ifupdown in Makefile fails on cross compile


AEberleMW:
Hi,

When cross-compiling mstpd (e.g. with [Yocto Project](https://www.yoctoproject.org)) the installation check in 
https://github.com/mstpd/mstpd/blob/f55f783610e651494736046605632056c99a61a0/Makefile.am#L87-L94
could fail, if the host system does not have `ifupdown` installed.

This especially happens when I run the build inside a container which only has a minimal set of tools and ifupdown is missing.

How could I handle this? Could you explain what the intention of this check is? Why do you check for `$(sysconfdir)/network/if-pre-up.d` and then create it in `$(etcdest)/network/if-pre-up.d`?

Thanks,
Marco

commodo:
apologies for the slow response;
this project isn't really high-up on my todo-list;
but i still try to keep it working as best as i can

hmm; this was introduced via 1b1d9a64

the initial version of `Makefile.am` had:
```
	if [ -d $(etcdir)/network/if-pre-up.d ] ; then ln -sf $(mstpdir)/ifupdown.sh $(etcdir)/network/if-pre-up.d/mstpctl ; fi
	if [ -d $(etcdir)/network/if-post-down.d ] ; then ln -sf $(mstpdir)/ifupdown.sh $(etcdir)/network/if-post-down.d/mstpctl ; fi
```

which looks a little saner;

I wonder if @PaulSD wants to pitch-in?

If not, I guess we could revert/fix it;
i can try to make some tests these days;

PaulSD:
Why does this fail on your system?  If ifupdown is not installed, then `$(sysconfdir)/network/if-pre-up.d` should not exist, and it should simply skip this without failing...

The problem that this code is trying to solve is:
When building a Debian package, the target/installation directory of the `make` command is a temporary path, not the real system.  In other words, $(etcdest) is something like `<current working directory>/debian/.target/etc/`.
So, the original code was checking for the existence of `$(etcdest)/network/if-post-down.d`, which will never exist when building a Debian package since that path is under a temporary directory.  Hence the symlink was never created when building a Debian package.
The current code instead checks `$(sysconfdir)/network/if-pre-up.d`, which should be the real `/etc/network/if-pre-up.d`.  If that exists, then we assume that ifupdown is installed and create the relevant path in the temporary target/installation directory, then create the symlink within it.  If ifupdown is not installed, then we should skip all of this.

commodo:
[...]

PR #127, opened 2021-09-29
build: add option to enable installation of ifupdown scripts

commodo:
The current setup with the ifupdown scripts, is that they get installed if
the system that builds mstpd is a Debian system.

When building for Debian on another host, or cross-compiling via Yocto or
some other build-system, this can be problematic as the host-system can be
different, so we cannot rely on the fact that the ifupdown package is
present on the target system.

So, the option is to enable the installation of MSTPD's ifupdown scripts
via a configure flag.
Initially this was discussed that this should be a
`--disable-ifupdown-scripts`, but this seems problematic as the only place
where this is needed is for Debian/Ubuntu systems.

Also, ifupdown seems to be going away from Debian as default, so maybe the
a --enable-install-ifupdown-scripts options is better.

Fixes: #126

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -87,17 +87,18 @@ bashcompdest=$(DESTDIR)$(bashcompletiondir)
 systemdunitdest=$(DESTDIR)$(systemdunitdir)
 nmlibdest=$(DESTDIR)$(nmlibdir)
 
+if ENABLE_INSTALL_IFUPDOWN_SCRIPTS
+IFUPDOWN_SCRIPT_INSTALL_CMDS = \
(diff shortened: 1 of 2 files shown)


commodo:
notifying @AEberleMW @konradgraefe @PaulSD 

i'm open for other suggestions;
this is what looked reasonable to me;

once this gets discussed, resolved and merged, i can publish a release to lock this feature (in a more official form)

PaulSD (approved):
LGTM

commodo:
I was thinking about maybe removing the `utils/ifupdown.sh` altogether from the installation.
But maybe for now, this is sufficient.

We could probably tweak it later.

PR #130, opened 2022-02-01
Update debian/sid to latest 0.1.0

rubensfig:


rubensfig:
Leaving the PR as a suggestion, is there any way to cleanly apply the state from 0.1.0 into debian/sid?

commodo:
well, we could do a rebase perhaps ;

but debian/sid is meant to be used for packaging mstpd into the Debian APT feed;

it's not currently used; but we were planning to get there :)

PR #132, opened 2022-02-27
mstpctl: add showportparams cmd

aliforouzan:
as I need to print some information(including enabled, port-id, internal-port-cost, state, role, admin-edge-port) for all of my interface(about 45 interfaces) use of showport command doesn't optimal(45*6 mstpdctl binary exec in my script). with showportparams command that will be add in this MR we can probe over than one parameter for an interface.

these changes will be useful for any one with my problem.

### DIFF
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -1257,7 +1257,8 @@ static int cmd_showportparams(int argc, char *const *argv)
             if(0 == strcmp(argv[argc - count + i], cist_port_params[j].str))
             {
                 param_id[i] = cist_port_params[j].id;
-                err = 0; break;
+                err = 0;
+                break;
             }
         }
         if (err)
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -1238,6 +1238,58 @@ static int cmd_showport(int argc, char *const *argv)
     return r;
 }
 
+static int cmd_showportparams(int argc, char *const *argv)
+{
+    int r = 0;
+
+    int br_index = get_index(argv[1], "bridge");
+    if(0 > br_index)
+        return br_index;
+
+    int i, count = argc - 3;
+    param_id_t param_id[count];
+
+    for(i = 0; i < count; ++i)
+    {
+        int err = 1;
+        for(int j = 0; j < COUNT_OF(cist_port_params); ++j)
+        {
+            if(0 == strcmp(argv[argc - count + i], cist_port_params[j].str))
+            {
+                param_id[i] = cist_port_params[j].id;
+                err = 0; break;
+            }
+        }
+        if (err)
+        {
+            fprintf(stderr, "Can't find index for %s param. "
+                            "Not a valid CSTP param.\n", argv[argc - count + i]);
+            return -1;
+        }
+    }
+
+    do_arraystart_fmt();
+
+    for(i = 0; i < count; ++i)
+    {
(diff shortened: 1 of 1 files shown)


commodo:
thanks for the patch;

i will take a look

ctl_main.c
```diff
@@ -1238,6 +1238,58 @@ static int cmd_showport(int argc, char *const *argv)
...
 
+static int cmd_showportparams(int argc, char *const *argv)
+{
+    int r = 0;
+
+    int br_index = get_index(argv[1], "bridge");
+    if(0 > br_index)
+        return br_index;
+
+    int i, count = argc - 3;
```
commodo, line 1249: if i read this correctly, if argc == 3, then count = 0 and nothing happens;
should we treat the `count == 0 ` case?

ctl_main.c
```diff
@@ -1238,6 +1238,58 @@ static int cmd_showport(int argc, char *const *argv)
...
+    param_id_t param_id[count];
```
commodo, line 1250: this `    param_id_t param_id[count];`  static initialization of array if more C99 specific;

ctl_main.c
```diff
@@ -1238,6 +1238,58 @@ static int cmd_showport(int argc, char *const *argv)
...
+
+    for(i = 0; i < count; ++i)
+    {
+        int err = 1;
+        for(int j = 0; j < COUNT_OF(cist_port_params); ++j)
+        {
+            if(0 == strcmp(argv[argc - count + i], cist_port_params[j].str))
+            {
+                param_id[i] = cist_port_params[j].id;
+                err = 0; break;
```
commodo, line 1260: nitpick: `break` on next line

commodo:
rest looks good

ctl_main.c
for454: This command should have [at least 3 argument](https://github.com/for454/mstpd/blob/777fc3e5630b96d301cf744f30c830372c73652b/ctl_main.c#L2257) that it means argc is at least equal to 4.

command syntax is like this:
```
 showportparams <bridge> <port> <param>... [params]
```

ctl_main.c
for454: any way, we need initialize the array?

ctl_main.c
for454: thank you, fix in https://github.com/mstpd/mstpd/pull/132/commits/c52ed4ff948c2a44abfc6193ef0f129c366047e1

ctl_main.c
commodo: apologies for the slow reply;
i missed your replies and i am a bit overloaded;

so, typically in older C versions you would allocate a static array of max-values
so, you would add a value to the enum called `PARAM_MAX` and do `param_id_t param_id[PARAM_MAX]`

the difference is that `PARAM_MAX` is a constant and `count` is a run-time variable;
older versions of the C compiler (C90 and below) don't know how to handle `param_id[count]`, so what you would do with these compilers is to allocate a max-array;

the run-time overhead isn't too big to do ` param_id[PARAM_MAX]` vs ` param_id[count]`

commodo (approved):

commodo:
so, i just checked;

seems we have more C99 things that slipped through

so, this is fine;

apologies for being slow

too many things to do :)

PR #142, opened 2023-06-26
mstpctl-utils-functions.sh: fix shellcheck warnings

KanjiMonster:
Fix the following shellcheck warnings:

```
In utils/mstpctl-utils-functions.sh line 5:
  while [ x"${1+set}" = xset ]
          ^---------^ SC2268 (style): Avoid x-prefix in comparisons as it no longer serves a purpose.

Did you mean:
  while [ "${1+set}" = set ]

In utils/mstpctl-utils-functions.sh line 53:
    echo $i
         ^-- SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
    echo "$i"
```

### DIFF
--- a/utils/mstpctl-utils-functions.sh
+++ b/utils/mstpctl-utils-functions.sh
@@ -2,7 +2,7 @@
 
 mstpctl_parse_ports()
 {
-  while [ x"${1+set}" = xset ]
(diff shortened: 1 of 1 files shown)


KanjiMonster:
LGTM.

Issue #100, opened 2020-09-17
RSTP creates temporary loops


jacekkow:
I have installed mstpd on a Debian 10 system which has a bridge (interface name is, conveniently, `bridge`) with two bonds attached (`bond-sw1`=`enp9s0f0`+`enp9s0f1` and `bond-sw2`=`enp9s0f2`+`enp9s0f3`), which are connected to two different switches running RSTP.

Everything seems to work:
```
# mstpctl showbridge bridge
bridge CIST info
  enabled         yes
  bridge id       8.000.00:...:22
  designated root 5.000.C4:...:A8
  regional root   8.000.00:...:22
  root port       bond-sw2 (#2)
  path cost     1000       internal path cost   0
  max age       20         bridge max age       20
  forward delay 15         bridge forward delay 15
  tx hold count 6          max hops             20
  hello time    2          ageing time          300
  force protocol version     rstp
  time since topology change 967s
  topology change count      4
  topology change            no
  topology change port       bond-sw2
  last topology change port  bond-sw1
# cat /sys/class/net/bridge/bridge/stp_state
2
```
But when I reboot one of the switches (being the root in RSTP tree, connected to `bond-sw2`), once it starts system shows that for a short period of time there is a loop present!
```
[  840.471477] bridge: received packet on bond-sw2 with own address as source address (addr:00:..:22, vlan:0)
[  840.471605] bridge: received packet on bond-sw2 with own address as source address (addr:00:..:22, vlan:0)
[  840.471720] bridge: received packet on bond-sw2 with own address as source address (addr:00:..:22, vlan:0)
```
(full [dmesg.log](https://github.com/mstpd/mstpd/files/5242301/dmesg.log) is attached)

It seems that mstpd allowed both bond-sw1 and bond-sw2 to enter forwarding state at once. My understanding is that RSTP should prevent occurrence of such an issue.

Could you point me as to why it happened? Scenario passes (no loops visible) with kernel's RTP.

commodo:
A lot of link flapping is present in those dmest logs.
Maybe you would need to tune some of the forward delays?
```
  forward delay 15         bridge forward delay 15
```

commodo:
Any more input here?
Will close this for now.

Feel free to re-open if issue still exists.

Generally speaking, if the link flags (as it is in the log), some temporary loops may occur.
Now, I have no idea why this happens here (in this log)

ErikHons:
[...]

PR #144, opened 2023-12-01
Refactor timers and fix temporary loops

ErikHons:
The concept here is to replace the "tick" based port timers with timerfds so that the changes they drive occur close to actual seconds when network peers would expect them. In particular, the auto edge transition could occur in only two seconds an occasionally beat a peers hello time resulting in a temporary loop.

This likely fixes issue #100


KanjiMonster:
This is definitely interesting, and I am pretty sure we have seen these loops in our internal testing.

Any reason this is just a draft? Any known issues here? I'm not very versed with epoll timers, does it prevent running expired timers in parallel, or would we need locking here?

ErikHons:
> Any reason this is just a draft? 
This fixes a specific issue, but the change probably makes sense for the project only if we do it comprehensively; replace all the tick-based timers with timerfds. I would like to get alignment on that and how to proceed before making an official PR.

KanjiMonster:
Looking at the kernel STP implementation, at it uses various timers, both per bridge as well per port, so I definitely think that using timers instead of ticks is the right way to go.

We just need to make sure that we avoid handling them in parallel and use locking where appropriate. The kernel uses one lock per bridge, so we might be fine with that as well.

KanjiMonster:
I would prefer everything in a single PR. Feel free to reuse this one, so the existing discussion stays linked.

And one commit per tick event conversion. Assuming we need locking to prevent concurrent accesses to data structures, probably add this first.

This should allow testing each conversion step individually, while keeping the changes together in a coherent topic.

PR #147, opened 2024-08-11
Update packet.c

Siva-Ranjan:


Siva-Ranjan (commented):
Added a printf statement in packet.c file

packet.c
```diff
@@ -131,7 +131,7 @@ static void packet_rcv(uint32_t events, struct epoll_event_handler *h)
 
     dump_packet(buf, cc);
 #endif
-
+    printf("sll_ifindex = %d and cc = %d\n", sl.sll_ifindex, cc);
```
commodo, line 134: so, if you build with the `PACKET_DEBUG` define, will it be sufficient for your use-case?

PR #172, opened 2026-03-23
Fix race condition on add_bridge command

thbtcllt:
This commit fixes a race condition where the list of interfaces given on add_bridge command is incorrect. Between the moment where the list of interfaces belonging to a bridge and the moment where mstpd receives and manages the message the list of interfaces belonging to the bridge can change. That leads mstpd to send BPDU packets on incorrect list of interfaces that can lead to misbehavior of the protocol. To fix the issue the list of interfaces belonging to the bridge is get by the mstpd daemon and not by the client (mstpcli  or other).

### DIFF
--- a/bridge_track.c
+++ b/bridge_track.c
@@ -130,6 +130,7 @@ static port_t * find_if(bridge_t * br, int if_index)
 
 static inline void delete_if(port_t *prt)
 {
+    INFO("Del iface %s", prt->sysdeps.name);
     MSTP_IN_delete_port(prt);
     free(prt);
 }
@@ -161,7 +162,10 @@ void bridge_one_second(void)
 {
     bridge_t *br;
     list_for_each_entry(br, &bridges, list)
-        MSTP_IN_one_second(br);
+    {
+        if(br->stp_enabled)
+            MSTP_IN_one_second(br);
+    }
 }
 
 /* New MAC address is stored in addr, which also holds the old value on entry.
@@ -203,7 +207,7 @@ static void set_br_up(bridge_t * br, bool up)
         MSTP_IN_set_bridge_address(br, br->sysdeps.macaddr);
     }
 
-    if(changed)
+    if(changed && br->stp_enabled)
         MSTP_IN_set_bridge_enable(br, br->sysdeps.up);
 }
 
@@ -258,7 +262,7 @@ static void set_if_up(port_t *prt, bool up)
             changed = true;
         }
     }
-    if(changed)
+    if(changed && prt->bridge->stp_enabled)
         MSTP_IN_set_port_enable(prt, prt->sysdeps.up, prt->sysdeps.speed,
                                 prt->sysdeps.duplex);
 }
@@ -277,7 +281,15 @@ int bridge_notify(int br_index, int if_index, bool newlink, unsigned flags)
     if((br_index >= 0) && (br_index != if_index))
     {
         if(!(br = find_br(br_index)))
-            return -2; /* bridge not in list */
+        {
+            /* Auto-create bridge in monitoring mode */
+            if(!(br = create_br(br_index)))
+            {
+                ERROR("Couldn't create data for bridge interface %d", br_index);
+                return -2;
+            }
+            INFO("Auto-monitor bridge %s", br->sysdeps.name);
+        }
         int br_flags = get_flags(br->sysdeps.name);
         if(br_flags >= 0)
             set_br_up(br, !!(br_flags & IFF_UP));
@@ -339,7 +351,15 @@ int bridge_notify(int br_index, int if_index, bool newlink, unsigned flags)
             if(br_index == if_index)
             {
                 if(!(br = find_br(br_index)))
-                    return -2; /* bridge not in list */
+                {
+                    /* Auto-create bridge in monitoring mode */
+                    if(!(br = create_br(br_index)))
+                    {
+                        ERROR("Couldn't create data for bridge interface %d", br_index);
+                        return -2;
+                    }
+                    INFO("Auto-monitor bridge %s", br->sysdeps.name);
+                }
                 set_br_up(br, up);
             }
         }
@@ -835,14 +855,11 @@ int CTL_set_fids2mstids(int br_index, __u16 *fids2mstids)
     return MSTP_IN_set_all_fids2mstids(br, fids2mstids) ? 0 : -1;
 }
 
-int CTL_add_bridges(int *br_array, int* *ifaces_lists)
+int CTL_add_bridges(int *br_array)
 {
-    int i, j, ifcount, brcount = br_array[0];
-    bridge_t *br, *other_br;
-    port_t *prt, *nxt;
-    int br_flags, if_flags;
-    int *if_array;
-    bool found;
+    int i, brcount = br_array[0];
+    bridge_t *br;
+    port_t *prt;
 
     for(i = 1; i <= brcount; ++i)
     {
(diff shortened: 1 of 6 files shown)


KanjiMonster:
Thank you for the PR

> The two fixes are independent and I can create separate pull requests if needed.

Please do so, and please sign off your commits. I do have comments and opinions, but let's first split this per-issue ;-)

thbtcllt:
The new version does the suggested modifications

bridge_ctl.h
```diff
@@ -38,6 +38,7 @@ typedef struct
     char name[IFNAMSIZ];
 
     bool up;
+    bool stp_managed;
```
KanjiMonster: I don't think we need a second stp_managed in `sysdep_br_data_t`, once in `bridge_t` should be enough.

bridge_track.c
```diff
@@ -161,7 +162,10 @@ void bridge_one_second(void)
 {
     bridge_t *br;
     list_for_each_entry(br, &bridges, list)
-        MSTP_IN_one_second(br);
+    {
+        if(br->sysdeps.stp_managed)
```
KanjiMonster:
```suggestion
        if(br->stp_managed)
```

bridge_track.c
```diff
@@ -203,7 +207,7 @@ static void set_br_up(bridge_t * br, bool up)
         MSTP_IN_set_bridge_address(br, br->sysdeps.macaddr);
     }
 
-    if(changed)
+    if(changed && br->sysdeps.stp_managed)
```
KanjiMonster:
```suggestion
    if(changed && br->stp_managed)
```

bridge_track.c
```diff
@@ -277,7 +281,12 @@ int bridge_notify(int br_index, int if_index, bool newlink, unsigned flags)
     if((br_index >= 0) && (br_index != if_index))
     {
         if(!(br = find_br(br_index)))
-            return -2; /* bridge not in list */
+        {
+            /* Auto-create bridge in monitoring mode */
+            if(!(br = create_br(br_index)))
+                return -2;
```
KanjiMonster, line 289: You should print an error if `create_br()` fails, e.g. the one from `CTL_add_bridge()`
```suggestion
            if(!(br = create_br(br_index))) {
                ERROR("Couldn't create data for bridge interface %d", br_index);
                return -2;
            }
```

bridge_track.c
```diff
@@ -318,7 +327,8 @@ int bridge_notify(int br_index, int if_index, bool newlink, unsigned flags)
             delete_if(prt);
             return 0;
         }
-        set_if_up(prt, running); /* And speed and duplex */
+        if(br->sysdeps.stp_managed)
```
KanjiMonster: This check belongs into `set_if_up()` (we still should update `sysdeps.{up, speed, duplex}`).

bridge_track.c
```diff
@@ -672,6 +687,7 @@ int CTL_get_cist_bridge_status(int br_index, CIST_BridgeStatus *status,
 
     CTL_CHECK_BRIDGE;
     MSTP_IN_get_cist_bridge_status(br, status);
+    status->stp_managed = br->sysdeps.stp_managed;
```
KanjiMonster: Just use `bridge_t::stp_managed` and let `MSTP_IN_get_cist_bridge_status()` populate it.

bridge_track.c
```diff
@@ -854,50 +869,22 @@ int CTL_add_bridges(int *br_array, int* *ifaces_lists)
...
-                }
-            }
-            if(!found)
-                delete_if(prt);
-        }
-        /* add all new interfaces from the list */
-        for(j = 1; j <= ifcount; ++j)
+
+        if(br->sysdeps.stp_managed)
+            continue; /* already managed */
```
KanjiMonster:
```suggestion
        if(br->stp_managed)
            continue; /* already managed */
```

bridge_track.c
```diff
@@ -837,13 +853,12 @@ int CTL_set_fids2mstids(int br_index, __u16 *fids2mstids)
...
-    port_t *prt, *nxt;
-    int br_flags, if_flags;
-    int *if_array;
-    bool found;
+    int i, brcount = br_array[0];
+    bridge_t *br;
+    port_t *prt;
+    int if_flags;
 
+    (void) ifaces_lists;
```
KanjiMonster: Just drop the argument; we don't need it anymore.

bridge_track.c
```diff
@@ -854,50 +869,22 @@ int CTL_add_bridges(int *br_array, int* *ifaces_lists)
...
-            if(!found)
-                delete_if(prt);
-        }
-        /* add all new interfaces from the list */
-        for(j = 1; j <= ifcount; ++j)
+
+        if(br->sysdeps.stp_managed)
+            continue; /* already managed */
+
+        br->sysdeps.stp_managed = true;
```
KanjiMonster:
```suggestion
        br->stp_managed = true;
```

bridge_track.c
```diff
@@ -854,50 +869,22 @@ int CTL_add_bridges(int *br_array, int* *ifaces_lists)
...
-            }
-            if(NULL == (prt = create_if(br, if_array[j])))
-            {
-                INFO("Couldn't create data for interface %d (master %s)",
-                     if_array[j], br->sysdeps.name);
-                continue;
-            }
             if(0 <= (if_flags = get_flags(prt->sysdeps.name)))
                 set_if_up(prt, (IFF_UP | IFF_RUNNING) ==
                                (if_flags & (IFF_UP | IFF_RUNNING))
```
KanjiMonster: If we allow `set_if_up()` for non-managed bridges, then `sysdeps` will contain the current state, so we can just do
```suggestion
            MSTP_IN_set_port_enable(prt, prt->sysdeps.up, prt->sysdeps.speed, prt->sysdeps.duplex);
```

and skip retrieving the state from sysfs.

bridge_track.c
```diff
@@ -911,9 +898,24 @@ int CTL_add_bridges(int *br_array, int* *ifaces_lists)
...
+            continue;
+        if(br->sysdeps.stp_managed)
+        {
+            INFO("Disable STP on bridge %s", br->sysdeps.name);
+            /* Disable all ports */
+            list_for_each_entry(prt, &br->ports, br_list)
+                MSTP_IN_set_port_enable(prt, false, 0, 0);
+            /* Disable the bridge */
+            MSTP_IN_set_bridge_enable(br, false);
+            br->sysdeps.stp_managed = false;
```
KanjiMonster:
```suggestion
            br->stp_managed = false;
```

ctl_main.c
```diff
@@ -113,6 +113,7 @@ static inline void do_arraynext_fmt(void)
 typedef enum {
     PARAM_NULL = 0,
     /* bridge params */
+    PARAM_STPMANAGED,
```
KanjiMonster: Please move this to the bottom of the (bridge) enum values, as a separation from the standard defined variables.

ctl_main.c
```diff
@@ -181,6 +182,7 @@ typedef struct {
 } cmd_param_t;
 
 static const cmd_param_t cist_bridge_params[] = {
+    { PARAM_STPMANAGED,      "stp managed" },
```
KanjiMonster: likewise, also please align the indentation.

ctl_main.c
```diff
@@ -213,7 +215,8 @@ static int do_showbridge_fmt_plain(const CIST_BridgeStatus *s,
     {
         case PARAM_NULL:
             printf("%s CIST info\n", br_name);
-            printf("  enabled         %s\n", BOOL_STR(s->enabled));
+            printf("  stp managed        %s\n", BOOL_STR(s->stp_managed));
+            printf("  enabled         %-5s", BOOL_STR(s->enabled));
```
KanjiMonster: Did you intend to switch those two? Now `enabled` will be joined in one line with `bridge id`.

mstp.h
```diff
@@ -611,6 +611,7 @@ typedef struct
     bridge_identifier_t regional_root;
     unsigned int internal_path_cost;
     bool enabled; /* not in standard */
+    bool stp_managed; /* not in standard */
```
KanjiMonster: Maybe call this `stp_enabled`, mirroring the `enabled` above? This would somewhat mirror the MIB variable `ieee8021SpanningTreePortEnable` (which does not actually appear anywhere else except in `IEEE8021-SPANNING-TREE` - must be a leftover from a very old standard).

KanjiMonster (changes requested):
Looking much better. I think we don't need two variables to store this, any stp-managed does IMHO not convey clearly what this means, so I suggest renaming it to stp-enabled.

thbtcllt:
The new version normally fixes all the remarks:
- stp_managed is renamed stp_enabled.
- stp_enabled is added in bridge_t structure and no more in sysdep_br_data_t structure.
- set_if_up is always call on netlink message reception to set speed, flags .... Call to MSTP_IN_set_port_enable is called if a  change is detected and stp_enabled set to true
- addbridge API is simplified (remove all reference to iface list) and is now similar to dellbridge. For compatibility with tools that reimplement mstpctl receiving trailing bytes for addbridge is accepted.
- output of showbridge is fixed
- Fix some incorrect indentation

KanjiMonster (approved):
The code changes now look good, could you please squash the two commits into one? Then I'll merge it.

thbtcllt:
I have squashed the two commits and rebased it on the head of the master branch.

PR #174, opened 2026-03-26
replace custom dont_txmt and bpdu_filter_port with enableBPDU{rx,tx}

KanjiMonster:
Replace custom `donttxmt` and `bpdufilterport` options with standard `enableBPDUtx` and `enableBPDUtx`, as defined in the optional L2GP extension for MSTP.

While we don't implement L2GP that shouldn't stop is from supporting parts of it.

To not break existing setups, still allow setting the old options, but translate this to setting the appropriate standard option to the inverse value.

This also indirectly fixes the issue that there was no way of retrieving the current configuration of `donttxmt`, as the current value of `enableBPDUtx` is now available.

### DIFF
--- a/ctl_main.c
+++ b/ctl_main.c
@@ -155,6 +155,7 @@ typedef enum {
     PARAM_RESTRTCN,
     PARAM_PORTHELLOTIME,
     PARAM_DISPUTED,
+    PARAM_ENABLEBPDURX,
     PARAM_BPDUGUARDPORT,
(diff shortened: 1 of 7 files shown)


KanjiMonster:
Noticed that I also should update the man-files and support scripts, so I did.

commodo (approved):

KanjiMonster:
Drafted until I added unit tests for it.

PR #178, opened 2026-04-22
ubus: allow dynamic bridge stp configuration over ubus

goemansb:
Add optional ubus (OpenWrt micro bus architecture) support to enable dynamic bridge configuration via ubus IPC.

A few years ago, a fork of **mstpd** named **ustp** was created to be used in **OpenWrt**. This fork got rid of the **mstpctl** configuration tool and instead provided an API over **ubus** to dynamically set (R)STP configuration. This PR is an attempt to port the **ubus** integration from **ustp** back to **mstpd** (albeit with a slightly different API) while still keeping the **mstpctl** tool available. I have been in contact with the original developer of **ustp**, Felix Fietkau, and he gave permission to try and port some of his work. 

Since there are probably many use cases where the **ubus** interface provides no extra value, I have made it into an optional compile-time feature (enabled by the **--enable-ubus** flag). If you compile with this option enabled, an **mstpd** **ubus** object will be registered providing two methods: **add_bridge** and **delete_bridge**. These methods function similar to respectively the **addbridge** and **delbridge** commands from **mstpctl** with the exception that **add_bridge** takes a number of optional arguments (proto, forward_delay, max_age, ageing_time) to (re)configure STP parameters.

I have left the original **ustp** copyright headers in some of the files since I didn't know what to do with them. I can change them if needed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,11 +5,42 @@ on: [ push, pull_request ]
 jobs:
   linux:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        config:
+          - name: "default"
+            flags: "--enable-devel"
+            install_ubus: false
+          - name: "with-ubus"
+            flags: "--enable-devel --enable-ubus"
+            install_ubus: true
     steps:
-    - uses: actions/checkout@v2
-    - name: build
+    - uses: actions/checkout@v6
+    - name: Install ubus dependencies
+      if: matrix.config.install_ubus
       shell: bash
-      run: ./autogen.sh && ./configure --enable-devel && make V=s
+      run: |
+        sudo apt update
+        sudo apt install -y libjson-c-dev cmake liblua5.1-0-dev lua5.1
+
+        # Build libubox
+        git clone https://git.openwrt.org/project/libubox.git
+        cd libubox
+        cmake .
+        make
+        sudo make install
+        cd ..
+
+        # Build libubus
+        git clone https://git.openwrt.org/project/ubus.git
+        cd ubus
+        cmake .
+        make
+        sudo make install
+        sudo ldconfig
+    - name: build (${{ matrix.config.name }})
+      shell: bash
+      run: ./autogen.sh && ./configure ${{ matrix.config.flags }} && make V=s
     - name: shellcheck
       shell: bash
       run: shellcheck utils/ifupdown.sh.in utils/mstp_config_bridge.in utils/mstpctl-utils-functions.sh
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,6 @@
 ACLOCAL_AMFLAGS = -I m4
 
-sbin_PROGRAMS = mstpd mstpctl
+sbin_PROGRAMS = mstpd
 
 mstpd_SOURCES = \
 	main.c epoll_loop.c epoll_loop.h clock_gettime.h brmon.c \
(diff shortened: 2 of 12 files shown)


commodo:
Oh this brings back memories.

I wrote some UCI stuff for mstpd a while back (which was wrapping mstpctl), but did not get a chance to upstream it.

I will leave it up to  @KanjiMonster  for his thoughts/decisions.
I have no strong opinions about ubus in mstpd;  it just makes me nostalgic.
Pinging @nbd168 for info/thoughts.

Thanks
Alex

KanjiMonster:
Just some quick thoughts for now:

- I'm pretty sure this will conflict with https://github.com/mstpd/mstpd/pull/172, which I want to merge first
- ubus vs. mstpctl should probably be exclusive, not both at the same time (one source of truth for configuration). Can be either at build time or at run time.
- it's OpenWrt, not OpenWRT
- no genAI generated code please

KanjiMonster:
Oh, and it should be at least build-tested by the CI, to avoid accidental breakage in the future

AlbrechtL:
> ubus vs. mstpctl should probably be exclusive, not both at the same time (one source of truth for configuration). Can be either at build time or at run time.

@KanjiMonster I think we need mstpctl on the OpenWrt device because it gives important information which you can't get elsewhere.

Here a real world example. E.g. the "designated root" information or root port information is useful for network troubleshooting.
```
root@OpenWrt:~# mstpctl showbridge switch
switch CIST info
  stp enabled     yes
  enabled         yes
  bridge id       8.000.50:E0:39:F4:CF:7A
  designated root 8.000.00:15:7E:1D:E6:58
  regional root   8.000.50:E0:39:F4:CF:7A
  root port       lan1 (#1)
  path cost     20000      internal path cost   0
  max age       20         bridge max age       20
  forward delay 15         bridge forward delay 15
  tx hold count 6          max hops             20
  hello time    2          ageing time          300
  force protocol version     rstp
  time since topology change 204
  topology change count      1
  topology change            no
  topology change port       None
  last topology change port  lan1
root@OpenWrt:~#
```

mundschenk-at:
Not necessarily. You can just add a ubus method to get this information instead.

AlbrechtL:
Agree. But this PR and well as my PoC PR https://github.com/mstpd/mstpd/pull/181 are adding the ubus functionality to `msptd`.

To me it looks like adding the most or all `mstpctl` functions via ubus in into `msptd` doesn't makes sense. Would it be an option to leave `msptd` untouched and enhance `mstpctl` instead? I mean adding ubus and a daemon option to `mstpctl` would do the job, or I'm wrong? The disadvantage would be that on OpenWrt systems two daemons ( `msptd` and `mstpctl`) are running.

goemansb:
> Oh, and it should be at least build-tested by the CI, to avoid accidental breakage in the future

I've added a second job in the CI for this. I've also made an attempt to address your other initial remarks. Right now, ubus support and mstpctl are mutually exclusive but curious to know what you think about the comment from @AlbrechtL above.

AlbrechtL:
In https://github.com/mstpd/mstpd/issues/180#issuecomment-4363443425 I posted a different idea and a proof of concept of a tool that acts as a bridge or adapter between ubus and `mstpctrl`. With this tool we leave `mstpd` and `mstpctrl` unchanged.

I needed it to add more ubus methods to build an experiments rstp-luci-app. Here is a screenshot of roughly working prototype ([code](https://github.com/AlbrechtL/openwrt-luci/tree/switch-luci-apps/applications/luci-app-rstp)).
grafik

KanjiMonster:
> To me it looks like adding the most or all `mstpctl` functions via ubus in into `msptd` doesn't makes sense. Would it be an option to leave `msptd` untouched and enhance `mstpctl` instead? I mean adding ubus and a daemon option to `mstpctl` would do the job, or I'm wrong? The disadvantage would be that on OpenWrt systems two daemons ( `msptd` and `mstpctl`) are running.

After thinking about this is a bit more, I relaxed my requirement a bit and think best would be to have ubus as a runtime option for the daemon, so if you enable it it registers via ubus etc.

And my thought is that if ubus mode is enabled, you need to use ubus to configure the bridge and ports. So mstpd would then refuse any SET calls from mstpctl, but still allow any GET calls (to make it easy to view the configuration, but also to for debugging etc).

AlbrechtL:
@KanjiMonster This sounds like a plan!

@goemansb Is it possible for you to work on that?

I can offer testing. But for coding I can only offer AI based coding which may not accepted here.

PR #179, opened 2026-04-28
Fix issue with a configuration name of 32 characters

thbtcllt:
Since version 0.0.9 and commit d27d7e93485d (treewide: allow room for null-char in strncpy() ) the last character of a configuration name of 32 characters is always lost and replaced by a NULL character.
As far as I know the standard does not have this requirement the configuration name in the MSTP frame can be non NULL terminated.

### DIFF
--- a/mstp.c
+++ b/mstp.c
@@ -1629,8 +1629,13 @@ void MSTP_IN_set_mst_config_id(bridge_t *br, __u16 revision, __u8 *name)
         assign(br->MstConfigId.s.revision_level, valueRevision);
         memset(br->MstConfigId.s.configuration_name, 0,
                sizeof(br->MstConfigId.s.configuration_name));
+        /* configuration_name can be non NUL terminated cf 802.1Q-2022 section 13.8
+         * copy everything and ignore GCC warning */
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
         strncpy((char *)br->MstConfigId.s.configuration_name, (char *)name,
-                sizeof(br->MstConfigId.s.configuration_name) - 1);
+                sizeof(br->MstConfigId.s.configuration_name));
+#pragma GCC diagnostic pop
         br_state_machines_begin(br);
     }
 }


mstp.c
```diff
@@ -1629,8 +1629,8 @@ void MSTP_IN_set_mst_config_id(bridge_t *br, __u16 revision, __u8 *name)
         assign(br->MstConfigId.s.revision_level, valueRevision);
         memset(br->MstConfigId.s.configuration_name, 0,
                sizeof(br->MstConfigId.s.configuration_name));
-        strncpy((char *)br->MstConfigId.s.configuration_name, (char *)name,
-                sizeof(br->MstConfigId.s.configuration_name) - 1);
+        size_t n = strnlen((char *)name, sizeof(br->MstConfigId.s.configuration_name));
+        memcpy(br->MstConfigId.s.configuration_name, name, n);
```
KanjiMonster: Since `strncpy()` does not add a NULL terminator if the source string is as long or longer than the destination buffer, wouldn't it be enough to do:

```suggestion
        strncpy((char *)br->MstConfigId.s.configuration_name, (char *)name,
                sizeof(br->MstConfigId.s.configuration_name));
```

KanjiMonster (changes requested):
The change looks reasonable, but since then `configuration_name` may not be NULL-terminated, we need to be extra careful every time we print it. Did you verify all places have appropriate handling?

Also the appropriate section (in 802.1Q-2022) is in "13.8 MST Configuration Identifier (MCID)":

> Each MCID contains the following components:
> ...
> b) The Configuration Name, a variable length text string encoded within a fixed field of 32 octets, conforming to IETF RFC 2271’s definition of SnmpAdminString. If the Configuration Name is less than 32 characters, the text string should be terminated by the NUL character, with the remainder of the 32-octet field filled with NUL characters. Otherwise, the text string is encoded with no terminating NUL character.

Pleas add that to the commit message.

mstp.c
thbtcllt: The use of strncpy leads to the following warning:
mstp.c: In function ‘MSTP_IN_set_mst_config_id’:
mstp.c:1620:9: warning: ‘__builtin_strncpy’ specified bound 32 equals destination size [-Wstringop-truncation]
 1620 |         strncpy((char *)br->MstConfigId.s.configuration_name, (char *)name,

as the NUL character can not be always set.
That is the reason I use memcpy with the size of the string limited to 32.
Same result with no warning.

thbtcllt:
> Did you verify all places have appropriate handling?

Yes I have checked and tested this point.
The configuration name is only printed by the showmstconfid command and protection again non NUL terminated string is done with the %.*s option:

printf("  Configuration Name:   %.*s\n", CONFIGURATION_NAME_LEN,  cfgid->s.configuration_name);

mstp.c
KanjiMonster: Let's not make the code worse just to avoid a warning.

Using https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html it should be possible to disable the warning just for that `strncpy()` call.

KanjiMonster:
> > Did you verify all places have appropriate handling?
> 
> Yes I have checked and tested this point. The configuration name is only printed by the showmstconfid command and protection again non NUL terminated string is done with the %.*s option:
> 
> printf(" Configuration Name: %.*s\n", CONFIGURATION_NAME_LEN, cfgid->s.configuration_name);

Then please do include that information in the commit message, then I don't need to ask ;-)

thbtcllt:
I have updated the commit log to indicate more clearly that print of non NUL name is tested (my previous sentence: "The 'mstpctl showmstconfid br0' command already manages this case" was too elliptic)
I use the pragma to remove the warning even if I am not a fan of that, it is too often a lazy solution that can mask real issue.

KanjiMonster:
> I use the pragma to remove the warning even if I am not a fan of that, it is too often a lazy solution that can mask real issue.

I agree with you in general, but I think here it is warranted, since this is the intended behavior, and not an actual issue. Also it's only disabled for the one line, and you added (as I also wanted to suggest, but forgot) a comment explaining why it is okay here.

Maybe in the future GCC becomes clever enough to detect open coding `strncpy()` and starts warning about it as well, so I think being explicit about this not being an issue is good.

KanjiMonster (approved):
LGTM now

PR #183, opened 2026-05-06
utils: fix bash completion to use ip instead of brctl

nadzyah:
I'm packaging mstpd for Debian as part of the Debian Networking team. While working on that, I've noticed that using `brctl` in the bash completion script causes a `command not found` error when completing bridge names on systems without `bridge-utils` installed. As the program already depends on `iproute2` itself (required by bridge-stp at runtime), I've decided to replace it with `ip -br` call to avoid adding a new runtime dependency

### DIFF
--- a/utils/bash_completion
+++ b/utils/bash_completion
@@ -24,8 +24,8 @@ _mstpctl()
                 debuglevel|showall)
                     ;;
                 *)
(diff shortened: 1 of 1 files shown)


commodo (approved):

KanjiMonster (approved):
LGTM, and works as advertised :)

PR #184, opened 2026-05-06
Add bridge-stp.8, mstp_restart.8, and mstpd.8 man pages

nadzyah:
As I was packaging mstpd for Debian, I wrote the man pages for the `bridge-stp`, `mstp_restart`, and `mstpd` binaries. I would appreciate it if you could review them, provide your suggestions or feedback, and, if possible, include these man pages in the code for the next release. This would help reduce the maintenance burden

### DIFF
--- a/utils/bridge-stp.8
+++ b/utils/bridge-stp.8
@@ -1,0 +1,239 @@
+.Dd April 28, 2026
+.Dt BRIDGE-STP 8
+.Os
+.Sh NAME
+.Nm bridge-stp
+.Nd kernel helper script that drives mstpd for Linux bridges
+.Sh SYNOPSIS
+.Nm
+.Ar bridge
+.Cm start | stop
+.Nm
+.Cm restart | restart_config
+.Sh DESCRIPTION
+.Nm
+is the kernel STP helper script for
+.Xr mstpd 8 .
+The Linux kernel invokes
+.Pa /sbin/bridge-stp
+.Ar bridge Cm start
+when STP is enabled on a bridge (via
+.Ic brctl stp Ar bridge Cm on
+or
+.Ic ip link set Ar bridge Ic type bridge stp_state 1 ) ,
+and
+.Pa /sbin/bridge-stp
+.Ar bridge Cm stop
+when STP is disabled.
+On most distributions
+.Pa /sbin
+is a symbolic link to
+.Pa /usr/sbin ,
+so the kernel resolves
+.Pa /sbin/bridge-stp
+to the script installed under
+.Pa /usr/sbin .
+If the script exits
+.Cm 0 ,
+the kernel switches the bridge into
+.Em user_stp
+mode and lets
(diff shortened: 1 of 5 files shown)


utils/bridge-stp.8
```diff
@@ -0,0 +1,228 @@
...
+.Ar bridge
+.Cm start | stop
+.Nm
+.Cm restart | restart_config
+.Sh DESCRIPTION
+.Nm
+is the kernel STP helper script for
+.Xr mstpd 8 .
+The Linux kernel runs
+.Pa /usr/sbin/bridge-stp
```
KanjiMonster: The installation location depends on `exec-prefix`/`sbindir` passed to configure. On systems where `/sbin` is symlink to `/usr/sbin` this is true, but on systems where it isn't `bridge-stp` is in `/sbin`, not `/usr/sbin`.

The kernel actually tries to execute `/sbin/bridge-stp` (which may get resolved to `/usr/sbin/bridge-stp`).

I have no idea if this can be set automatically based on the configuration.

utils/bridge-stp.8
```diff
@@ -0,0 +1,228 @@
...
+.El
+.Sh FILES
+.Bl -tag -width Ds
+.It Pa /usr/sbin/bridge-stp
+The script itself.
+The kernel invokes it as
+.Pa /sbin/bridge-stp ,
+which on Debian resolves to this path via the
+.Pa /sbin No \(-> Pa /usr/sbin
+compatibility symlink.
```
KanjiMonster, line 174: .. wait, here you even explain that. I think that explanation would make more sense at the top.

Also can we replace Debian with "many distributions" or so? While most distributions do that, not everyone does (e.g. OpenWrt doesn't).

KanjiMonster (changes requested):
Mostly looks good, just some minor nit-picks

utils/bridge-stp.8
nadzyah: I've initially written this manpage for Debian and then decided to forward it :D 
Debian policy forbids installing binaries directly under `/sbin` dir: https://www.debian.org/doc/debian-policy/ch-files.html#binaries
> That is, packages must not install files to paths matching the glob patterns `/bin/*`, `/lib/*`, `/lib*/*` and `/sbin/*`

So I specified the `/urs` prefix in the package building logic and indicated that path in the man page. I've updated it here to use just `/sbin`

utils/bridge-stp.8
nadzyah: Since this is a plain mdoc file (not processed by autoconf), I can't do `@sbindir@` substitution without adding build system plumbing. So I've documented both paths explicitly (the kernel-facing /sbin/bridge-stp in DESCRIPTION, and a note that it resolves to the installed location, which is typically `/sbin -> /usr/sbin`)

KanjiMonster (approved):
Looks good now

PR #186, opened 2026-05-06
debian: remove debian/ directory

nadzyah:
The package is maintained on Salsa as a non-native Debian package, so the debian/ directory belongs in the repository there: https://salsa.debian.org/networking-team/mstpd

### DIFF
--- a/debian/README.md
+++ b/debian/README.md
@@ -1,4 +1,0 @@
-# Packaging for Debian
-
-The packaging for Debian has been removed from this branch. It is now
-located in the `debian/sid` branch.


commodo:
So, then a question here.

Do we still need to keep/track a `debian/sid` branch anymore?

IIRC, I talked to a Debian maintainer several years ago, and the (back then mechanism) was that we keep a `debian/sid` branch.

nadzyah:
@commodo No, there is no need to maintain a `debian/sid` branch in the upstream repo. Packaging is maintained separately on Salsa as a non-native Debian package, which is the standard approach for projects that don't carry `debian/` in their upstream tree.

The bug report you may be thinking of ([#767013](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=767013)) actually shows the maintainer recommending exactly this:

> An alternative would be to do the official packaging on salsa.debian.org using traditional tools (notably, the released tarballs)

That is the approach we are using, so the debian/sid branch is no longer needed and can be dropped

commodo:
> @commodo No, there is no need to maintain a `debian/sid` branch in the upstream repo. Packaging is maintained separately on Salsa as a non-native Debian package, which is the standard approach for projects that don't carry `debian/` in their upstream tree.
> 
> The bug report you may be thinking of ([#767013](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=767013)) actually shows the maintainer recommending exactly this:
> 
> > An alternative would be to do the official packaging on salsa.debian.org using traditional tools (notably, the released tarballs)
> 
> That is the approach we are using, so the debian/sid branch is no longer needed and can be dropped

Oh wow.
Time has passed since that discussion.
I nearly forgot about that.

@KanjiMonster 
If no objections, let's delete this branch in the next weeks. unless needed for some other purpose.

commodo (approved):

KanjiMonster:
> @KanjiMonster If no objections, let's delete this branch in the next weeks. unless needed for some other purpose.

Yes, or maybe push a commit removing everything except for single README.md saying that this is now packaged via Debian's salsa and a link to it.

KanjiMonster (approved):

nadzyah:
> Yes, or maybe push a commit removing everything except for single README.md saying that this is now packaged via Debian's salsa and a link to it.

Could you please do this in the `debian/sid` dir and avoid having the `debian/` directory in the master branch and releases tarballs? Otherwise I have to remove its content manually, which is a bit extra work

KanjiMonster:
> > Yes, or maybe push a commit removing everything except for single README.md saying that this is now packaged via Debian's salsa and a link to it.
> 
> Could you please do this in the `debian/sid` dir and avoid having the `debian/` directory in the master branch and releases tarballs? Otherwise I have to remove its content manually, which is a bit extra work

What I meant is instead of deleting the debian/sid branch replacing its content with a single README.md, so anyone trying to update their debian/sid checkout will instantly see where they should go instead of getting an error.

Nothing will be added to master.

PR #187, opened 2026-05-07
libnetlink: sync with iproute2 v7.0.0

KanjiMonster:
Update our libnetlink copy to the version of iproute2 v7.0.0. This brings various enhancements:

* dynamic allocation of buffers instead of relying on callers to allocate them (and potentially overrun them)
* increased amount of accessors for netlink attributes
* type aware accessors for netlink attribute data
* per-object type filter functions
* etc

and a lot of fixes.

All users of the APIs were updated.

To fit into mstpd, the following changes were applied:

* dropped the include of `utils.h`, it was actually unused
* replaced `json_print.h` with stubs
* dropped the `#warning` for !LIBMNL (I don't want to bring new dependencies for now)

and the following changes reapplied:

* replaced all `fprintf(stderr, "foo\n")` with `ERROR("foo")`
* replaced all `perror("foo")` with `ERROR("foo: %m")`, fixing up several missing `%m` locations
* replaced all `exit(1)` with `return -1`, adding cleanup calls as necessary
* made `rtnl_listen()` return `0` on `EAGAIN`, instead of blocking

There is the potential of this breaking the build with old distributions with old kernel headers, if we find that we could start shipping a copy of kernel headers ourselves.

As a side effect this also fixes printing error strings returned from the kernel.

Before:

```
mstpd[837]: error, rtnl_talk: RTNETLINK answers
mstpd[837]: error, MSTP_OUT_set_state: swbridge:enp3s0f0 Couldn't set kernel bridge state blocking
```

After:

```
mstpd[780]: error, rtnl_talk_error: RTNETLINK answers: Network is down
mstpd[780]: error, MSTP_OUT_set_state: swbridge:enp3s0f0 Couldn't set kernel bridge state blocking
```

This should make debugging issues a bit easier.

### DIFF
--- a/bridge_track.c
+++ b/bridge_track.c
@@ -461,7 +461,7 @@ static int br_set_state(struct rtnl_handle *rth, unsigned ifindex, __u8 state)
 
     addattr8(&req.n, sizeof(req.buf), IFLA_PROTINFO, state);
 
-    return rtnl_talk(rth, &req.n, 0, 0, NULL, NULL, NULL);
+    return rtnl_talk(rth, &req.n, NULL);
 }
 
 static int br_flush_port(char *ifname)
--- a/brmon.c
+++ b/brmon.c
@@ -58,7 +58,7 @@ static struct epoll_event_handler br_handler;
 
 struct rtnl_handle rth_state;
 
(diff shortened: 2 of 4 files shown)


KanjiMonster:
Now runtime tested.

commodo (approved):

PR #189, opened 2026-05-08
bridge_track: skip setting STP states of down ports

KanjiMonster:
The kernel automatically sets ports going down to discarding/disabled, and ports going up to blocking.

While down, the kernel doesn't allow setting any STP states, so we should not attempt to do so.

This should be fine, since disabled ports should never be anything other than blocking/disabled, and freshly enabled ports start at blocking.

To ensure this assumption holds, add a check for STP state to be blocking or discarding for ports that are down.

Silences messages on startup like:

```
mstpd[780]: error, rtnl_talk_error: RTNETLINK answers: Network is down
mstpd[780]: error, MSTP_OUT_set_state: swbridge:enp3s0f0 Couldn't set kernel bridge state blocking
```

### DIFF
--- a/bridge_track.c
+++ b/bridge_track.c
@@ -517,14 +517,26 @@ void MSTP_OUT_set_state(per_tree_port_t *ptp, int new_state)
             state_name = "disabled";
             break;
     }
+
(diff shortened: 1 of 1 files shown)


vjardin:
LGTM

PR #198, opened 2026-09-18
brmon: use IFLA_INFO_KIND for bridge type detection

KanjiMonster:
Use `IFLA_LINKINFO`'s `IFLA_INFO_KIND` attribute to detect bridges. This avoids trying to guess based on the family and absence of `IFLA_MASTER`, and trying to read from sysfs.
    
Reading from sysfs will fail for deleted bridges, and bridges can have `IFLA_MASTER` set (e.g. VRFs). Using `IFLA_LINFO` avoids both.

This also allows us to drop `is_bridge()`, reducing the need for sysfs.


KanjiMonster:
Minor whitespace fix.

commodo (approved):

